PR #1, opened 2020-09-03
Migrate types from specs-actors

anorth:
This is a copy of most of the specs-actors `abi` package, in preparation for re-use across release branches.

No semantic change intended. Other notes:
- Lifted the `big` package out from under `abi`
- Split the `primitives.go` into `chain.go` and `actor.go`
- Does not bring the policy associated with the RegisteredProof types into this repo


abi/actor.go
```diff
@@ -0,0 +1,38 @@
...
+
+func (e MethodNum) String() string {
+	return strconv.FormatInt(int64(e), 10)
+}
+
+// Multiaddrs is a byte array representing a Libp2p MultiAddress
+type Multiaddrs = []byte
+
+// PeerID is a byte array representing a Libp2p PeerID
+type PeerID = []byte
```
anorth, line 37: I'm not so sure of these last two. I'd slightly prefer we dropped them and used `[]byte` directly. They both appear in miner info state, and hence in constructor params (which are also accessed by power actor).

abi/cid.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// When producing a CID for an IPLD block less than or equal to CIDInlineLimit
+	// bytes in length, the identity hash function will be used instead of
+	// HashFunction. This will effectively "inline" the block into the CID, allowing
+	// it to be extracted directly from the CID with no disk/network operations.
+	//
+	// This is currently -1 for "disabled".
+	//
+	// This is exposed for testing. Do not modify unless you know what you're doing.
+	CIDInlineLimit = -1
```
anorth, line 22: @Stebalien I think we should push this down to a member of the cidBuilder, like the codec. I refrained from doing that here.

gen/gen.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	gen "github.com/whyrusleeping/cbor-gen"
+
+	"github.com/filecoin-project/go-state-types/abi"
+)
+
+func main() {
+	// Common types
+	if err := gen.WriteTupleEncodersToFile("./abi/cbor_gen.go", "abi",
+		abi.PieceInfo{},
```
ZenGround0, line 12: Comparing with specs actors this is missing SectorID, adding before merge

abi/sector.go
```diff
@@ -0,0 +1,164 @@
...
+// SectorNumber is a numeric identifier for a sector. It is usually relative to a miner.
+type SectorNumber uint64
+
+func (s SectorNumber) String() string {
+	return strconv.FormatUint(uint64(s), 10)
+}
+
+// The maximum assignable sector number.
+// Raising this would require modifying our AMT implementation.
+//goland:noinspection GoUnusedConst
```
ZenGround0, line 22: Are these here on purpose?  I'd rather keep editor specific directives out of the source.  Planning to remove but we can revert later today.

abi/sector.go
```diff
@@ -0,0 +1,164 @@
...
+	RegisteredPoStProof_StackedDrgWinning64GiBV1  = RegisteredPoStProof(4)
+	RegisteredPoStProof_StackedDrgWindow2KiBV1    = RegisteredPoStProof(5)
+	RegisteredPoStProof_StackedDrgWindow8MiBV1    = RegisteredPoStProof(6)
+	RegisteredPoStProof_StackedDrgWindow512MiBV1  = RegisteredPoStProof(7)
+	RegisteredPoStProof_StackedDrgWindow32GiBV1   = RegisteredPoStProof(8)
+	RegisteredPoStProof_StackedDrgWindow64GiBV1   = RegisteredPoStProof(9)
+)
+
+// Metadata about a seal proof type.
+type SealProofInfo struct {
```
ZenGround0, line 99: Got confused about this at first but as I understand it this is part of not bringing registered proof policy in here. To integrate into specs actors we will have to wrap this inside specs_actors.SealProofInfo along with SectorMaxLifetime and WindowPoStPartitionSectors.

ZenGround0 (approved):

abi/cid.go
Stebalien: This is here to make it easy to play with. I'd just leave it. The builder is effectively just a singleton.

anorth: It's singleton until we want to change this, and then it's not.

PR #5, opened 2020-09-07
Add Empty type/value

anorth:
This comes from the `adt` (not `abi`) package in actors, but should really have been in `abi` since it features in many method signatures.

### DIFF
--- a/abi/empty.go
+++ b/abi/empty.go
@@ -1,0 +1,40 @@
+package abi
+
+import (
+	"fmt"
+	"io"
+
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+// The empty value represents absence of a value. It is used for parameter and return types for actor methods
+// that don't take/return any data. This saves a byte in serialization of messages and receipts: the serialized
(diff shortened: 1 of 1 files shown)


abi/empty.go
```diff
@@ -0,0 +1,40 @@
...
+// that don't take/return any data. This saves a byte in serialization of messages and receipts: the serialized
+// form is an empty byte slice, rather than a byte slice containing a single byte CBOR encoding of nil/empty/etc.
+//
+// The only expected use of this is as the type of a nil reference. Don't instantiate this type.
+//
+// This is primarily necessary due to Go's lack of a void type and our interface-based serialization scheme.
+type EmptyValue struct{}
+
+// A typed nil pointer to EmptyValue.
+//goland:noinspection GoUnusedGlobalVariable
```
ZenGround0, line 20: nit remove ide directive

ZenGround0 (approved):
LGTM

abi/empty.go
anorth: Done, but let's discuss policy here.

PR #8, opened 2020-09-09
Extract rt types

ZenGround0:
First step in the effort to make runtime interface changes less painful to upgrade.  Extracts CBORer and LogLevel.  I put these in the network package since the network version is already important for runtime consumers.  Give me a better package / file name suggestion and I'll implement it.


network/runtime.go
```diff
@@ -0,0 +1,35 @@
+package network
+
+import "io"
+
+// These interfaces are intended to match those from whyrusleeping/cbor-gen, such that code generated from that
+// system is automatically usable here (but not mandatory).
+type CBORMarshaler interface {
```
Stebalien, line 7: I'd actually just put these into a separate `cbor` package.

network/runtime.go
```diff
@@ -0,0 +1,35 @@
...
+	UnmarshalCBOR(r io.Reader) error
+}
+
+type CBORer interface {
+	CBORMarshaler
+	CBORUnmarshaler
+}
+
+// Specifies importance of message, LogLevel numbering is consistent with the uber-go/zap package.
+type LogLevel int
```
Stebalien, line 21: Maybe a new `log` package? Maybe just introduce a runtime package?  To reduce imports, we can type alias any of these inside specs-actors anyways.

network/runtime.go
anorth: SG. `cbor.Marshaler`, `cbor.Er` etc :)

network/runtime.go
anorth: `log` could be an annoying collision for consumers. `runtime` or `rt` could be ok, tho I would then wish that `ExitCode` was there too.

network/runtime.go
anorth: or `ser`, `serde`?

anorth (approved):

network/runtime.go
ZenGround0: Going with `rt`.  Will file an issue for moving exitcodes there as a nice to have if things are going really well early next week.

network/runtime.go
ZenGround0: Going with (Un)Marshaler and Er

PR #12, opened 2020-09-24
Update network versioning comments

arajasek:
Just throwing this up, we can merge it whenever feels appropriate


anorth:
Thanks for starting this. I have a similar WIP locally :-)

I'm also considering renaming the constants to symbolic names, probably matching the names Lotus has used internally. Let's discuss.

Stebalien:
New version in #15. I agree we need consistent naming (although I'd just use versions and drop the code names).

PR #13, opened 2020-09-26
Add a shared actor interface definition

Stebalien:
Ideally, this would just be defined in lotus (where the type is used).

However, I'd like to change the `BuiltinActors()` method in specs-actors to return `[]Actor`. This way, all actors for a specific specs-actors version can be registered with the lotus vm with a single `vm.Register(..., BuiltinActors()...)` call.

Unfortunately, the type `[]actors0.Actor` is not the same as `[]actors1.Actor`, even if `actors0.Actor` and `actors1.Actor` are identical interfaces. Hence the definition in this package.

### DIFF
--- a/rt/actor.go
+++ b/rt/actor.go
@@ -1,0 +1,33 @@
+package rt
+
+import (
+	"github.com/filecoin-project/go-state-types/cbor"
+	"github.com/ipfs/go-cid"
+)
+
+// VMActor is a concrete implementation of an actor, to be used by a Filecoin
+// VM.
+type VMActor interface {
+	// Exports returns a slice of methods exported by this actor, indexed by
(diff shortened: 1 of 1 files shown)


anorth (approved):

anorth:
We have multiple things called `Actor`, one notable other one being the name of the state structure in the top level of the state tree. Consider calling it `ActorCode` or `ActorExecutable` or something. Whatever you agree with another Lotus developer is probably fine (as is leaving it as-is).

Stebalien:
Good point. ActorCode and/or ActorType may be the most accurate.

Stebalien:
Ok, we've gone with "vmactor". It's not perfect, but we couldn't come up with a better name.

PR #15, opened 2020-10-01
fill in new network versions

Stebalien:
This patch adds a version 5 (pending, may change), and provides a "version count" for convenience.

### DIFF
--- a/network/version.go
+++ b/network/version.go
@@ -7,11 +7,13 @@ import "math"
 type Version uint
 
 const (
-	Version0 = Version(iota) // specs-actors v0.9.3
-	Version1                 // specs-actors v0.9.7
-	Version2                 // specs-actors v0.9.?
-	Version3                 // Coming soon
-	Version4                 // Who knows?
+	Version0 = Version(iota) // genesis   (specs-actors v0.9.3)
+	Version1                 // breeze    (specs-actors v0.9.7)
+	Version2                 // smoke     (specs-actors v0.9.8)
+	Version3                 // ignition  (specs-actors v0.9.11)
+	Version4                 // actors v2 (specs-actors v2.0.x (future))
+	Version5                 // sometime?
 
+	// VersionMax is the maximum version number
 	VersionMax = Version(math.MaxUint32)
 )


network/version.go
```diff
@@ -7,11 +7,16 @@ import "math"
...
-	Version1                 // specs-actors v0.9.7
-	Version2                 // specs-actors v0.9.?
-	Version3                 // Coming soon
-	Version4                 // Who knows?
+	Version0 = Version(iota) // 00000000: genesis   (specs-actors v0.9.3)
+	Version1                 // 00041280: breeze    (specs-actors v0.9.7)
+	Version2                 // 00051000: smoke     (specs-actors v0.9.)
+	Version3                 // 00094000: ignition  (specs-actors v0.9)
+	Version4                 // 00128888: actors v2 (specs-actors v2.0.x (future))
+	Version5                 // 00148888: liftoff   (specs-actors v2.0.x (future))
```
anorth, line 15: I don't think we should duplicate mainnet configuration like epoch numbers in here. These won't match future testnets etc.

Also I have already been burned once by describing a future version before it happened, and the description being then proven wrong.  At present, there is no intention of a version number change at liftoff.

Adding an opaque `Version5` is fine, but I'd retract all descriptions except for the specs-actors versions corresponding to actually releases that first included reference to those numbers.

network/version.go
```diff
@@ -7,11 +7,16 @@ import "math"
...
 
+	// VersionCount is the number of versions defined.
+	VersionCount
```
anorth, line 18: I omitted this intentionally because as soon as tried to use it I realised I'd set a trap for downstream things that might assume that `VersionCount` refers to a particular version, and break when new items were added here. Unless we major-version go-state-types every time we add a new one, I think this is unwise.

anorth (changes requested):

network/version.go
Stebalien: I added this so I could per-allocate an array of versions, but I never ended up using it. No reason to keep it.

network/version.go
Stebalien: > At present, there is no intention of a version number change at liftoff.

There will be migration code (to set a network name, at least). Does this not imply a network version number? I've been assuming that's what they're for.

Stebalien: > Adding an opaque Version5 is fine, but I'd retract all descriptions except for the specs-actors versions corresponding to actually releases that first included reference to those numbers.

What about the upgrade name? These versions are used for more than just specs-actors upgrades.

anorth: I believe there has already been a migration in code that did not change the version number - the great unburning and fund redistribution. The version number only *needs* to change when actor behaviour changes.

Stebalien: From what I can tell, that was the upgrade at "breeze" which bumped us from version 0 to version 1.

anorth:
Just a note from my comment in #12

> I'm also considering renaming the constants to symbolic names

The reason is that the numbers define an ordering that is not necessarily known at the time of writing code in the actors like `if rt.NetworkVersion >= Version4`. The author might be expecting version 4 to be "the one where we do X", but in fact another version might end up going in earlier for some emergency, or something in Lotus might bump the version number without Actors being involved. If we use symbolic names, we can re-order the constants and have abstract checks in the actor code.

E.g. we *intend* to have the actorsv2 upgrade ahead of liftoff, but something might derail that.

But we don't have to do it now.

anorth (approved):
I wouldn't label the future ones, but we can relabel later if necessary.

Stebalien:
> The reason is that the numbers define an ordering that is not necessarily known at the time of writing code in the actors like if rt.NetworkVersion >= Version4. The author might be expecting version 4 to be "the one where we do X", but in fact another version might end up going in earlier for some emergency, or something in Lotus might bump the version number without Actors being involved. Then the actors checks all need to be rewriten. If we use symbolic names, we can re-order the constants and have abstract checks in the actor code.

That makes a lot of sense, actually.

Stebalien:
I've removed version 5. If we'd like to version every upgrade, we can add it later (sometime before mainnet).

PR #16, opened 2020-10-13
Prep for upcoming upgrades

ZenGround0:
Should we speculatively add 7,8,9,10 so we can save a few minutes when developing upgrades under pressure in the next few weeks?


Stebalien:
Eh, I wouldn't add any versions until we need them honestly.

anorth (approved):

PR #24, opened 2021-03-04
make NextNotElapsed faster

Stebalien:
Previously, this would have to walk through all deadlines between the proving period start and the current epoch. That was fine for the use-case but I'd _like_ to be able to compute the next-not-elapsed deadline at a significant distance.

Obviously, this has a pretty exhaustive test case.

### DIFF
--- a/dline/deadline.go
+++ b/dline/deadline.go
@@ -72,11 +72,34 @@ func (d *Info) FaultCutoffPassed() bool {
 
 // Returns the next instance of this deadline that has not yet elapsed.
 func (d *Info) NextNotElapsed() *Info {
-	next := d
-	for next.HasElapsed() {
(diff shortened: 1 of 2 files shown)


dline/deadline.go
```diff
@@ -72,11 +72,34 @@ func (d *Info) FaultCutoffPassed() bool {
...
+	// handle negative period starts just in case.
+	if offset < 0 {
+		offset += d.WPoStProvingPeriod
+	}
+	// 2. determine the global period index.
+	globalPeriod := (d.CurrentEpoch / d.WPoStProvingPeriod)
+	// 3. Determine our next period start.
+	periodStart := globalPeriod*d.WPoStProvingPeriod + offset
+
+	// Backtrack so the period starts before the current epoch. This should usually run 0 or 1 times.
```
ZenGround0, line 92: Is it possible for this to run more than 1 times?  I don't see how.

ZenGround0 (approved):
Looks good

dline/deadline.go
Stebalien: It shouldn't unless I got something very wrong.

PR #28, opened 2021-05-28
add v14 and v15 reserved

lyswifter:
add v14 and v15 reserved for alleviate penalty window post


anorth (approved):
I'm approving that we need these reserved generally. This should not be taken as approval or not of the proposal to change Window PoSt penalties.

arajasek:
Stale

PR #29, opened 2021-07-08
Properly guard against 'hollow' objects

ribasushi:
These are often instantiated as part of a larger struct unmarshall
A test for this exact scenario is included - refer to it for exact scenario we are guarding against

This is a very fundamental library. **Do not merge without sign-offs from**
- @ZenGround0 
- @anorth 
- @Kubuxu 
- @Stebalien 

( PR from personal account due to lack of perms )

### DIFF
--- a/big/int_test.go
+++ b/big/int_test.go
@@ -81,6 +81,12 @@ func TestInt_MarshalUnmarshalJSON(t *testing.T) {
 	s, err := tnil.MarshalJSON()
 	require.NoError(t, err)
 	assert.Equal(t, "\"0\"", string(s))
+
+	parsedStruct := func() (s struct{ Big Int }) {
+		require.NoError(t, json.Unmarshal([]byte("{}"), &s))
+		return s
+	}
+	assert.Equal(t, parsedStruct().Big, Int{})
 }
 
 func TestOperations(t *testing.T) {
@@ -89,7 +95,7 @@ func TestOperations(t *testing.T) {
 		f        func(Int, Int) Int
 		expected Int
 	}{
-		{name: "Sum", f: Add, expected: NewInt(7000)},
+		{name: "Add", f: Add, expected: NewInt(7000)},
 		{name: "Sub", f: Sub, expected: NewInt(3000)},
 		{name: "Mul", f: Mul, expected: NewInt(10000000)},
 		{name: "Div", f: Div, expected: NewInt(2)},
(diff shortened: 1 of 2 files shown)


Stebalien:
In general, I'm fine treating the "nil" int as 0 (as is the go way...). This shouldn't affect consensus because, in cbor-gen:

1. We can never leave a BigInt as nil on decode.
2. We encode "nil" BigInts as 0 on encode.

I've actually run into this issue before because our genesis code used to leave some values as nil (although they were correctly turned into zeros when encoded on-chain).

However, I'd still prefer to reject empty/nil ints when decoding json (as we would in cbor).

ribasushi:
> However, I'd still prefer to reject empty/nil ints when decoding json

How would you do this when the entire map entry isn't there? In the test we decode a `"{}"`, there's nothing to reject... no?

Stebalien:
Ah, sorry, I misunderstood how encoding to JSON worked here (thought `{}` _was_ the zero type).

Yeah, handling the "missing" int as 0 is, unfortunately, the most reasonable choice for go.

ribasushi:
@Stebalien can haz ✅ then? 😸

Stebalien (approved):

big/int_test.go
```diff
@@ -116,6 +117,36 @@ func TestOperations(t *testing.T) {
 	assert.True(t, ta.Nil())
 }
 
+func TestInt_NilUnmarshal(t *testing.T) {
+
+	parsedStruct := func() (s struct{ Big Int }) {
+		require.NoError(t, json.Unmarshal([]byte("{}"), &s))
+		return s
+	}
```
anorth, line 125: This test is kinda confusing two different things. It would be more clear as two different tests:
- What do you get when you unmarshal nothing into an `Int`. Show exactly that you get a wrapped `nil`, separate from how that behaves.
- How do various operations deal with `nil` parameters. Just use an uninitialised `Int` without jumping through the hoops.

anorth (commented):
We need to be really clear with the argument about why this is ok to change without versioning the library, thus causing this change to be imported into the whole past blockchain validation as well as to different operating nodes at different times.

It's not enough to only talk about cbor-gen behaviour. An uninitialised Int could be created and operated on in some code path without involving serialization. Previously, that would panic, and now it won't. I think our argument why this is safe is that the panic could not possibly have been in the code path of a block validation that made it into the chain in the past, nor could in the future.

This is unfortunately a very difficult argument to make, for this or any other undefined behaviour in actors, because Lotus has taken a policy of deciding that undefined evaluation from actors should be memorialised on chain in perpetuity with a hard-coded exit code. See https://github.com/filecoin-project/lotus/pull/3697. At a minimum, we would have to revalidate the entire chain with this change to check the behaviour matches.

big/int_test.go
ribasushi: @anorth addressed in 0b9c5581fa140aba8fad28550b49584fb59d0b3b

ribasushi:
> We need to be really clear with the argument about why this is ok to change without versioning the library, thus causing this change to be imported into the whole past blockchain validation as well as to different operating nodes at different times.

Actually I have no solid opinion on what specs-actors should do here. My motivation for the change is that surface APIs are failing left and right due to missing payloads like illustrated in the test ( we use the very same library for any API call unmarshal ).

It would be conceivable to have specs-actors switch to its own fork maintaining the exact behavior like it does now.

ribasushi:
... we could **also** just version the library and fix every-but-one instance of https://pkg.go.dev/github.com/filecoin-project/go-state-types@v0.1.0/big?tab=importedby.

Stebalien:
> I think our argument why this is safe is that the panic could not possibly have been in the code path of a block validation that made it into the chain in the past, nor could in the future.

Hm. Yes, that's very annoying.

I'd be comfortable just auditing the _current_ code, but that's less than optimal and still quite a bit of work.

ribasushi:
@anorth @Stebalien @ZenGround0 am I reading right that this is what we want to do: https://github.com/filecoin-project/go-state-types/pull/29#issuecomment-877993578

I.e. switch lotus frontend to use a v2 or even a "go-filbigint-safe" or somesuch?

anorth:
> Actually I have no solid opinion on what specs-actors should do here.

I understand where you are coming from, but you can't change this library without having an opinion about actors. This library is logically part of actors, it's the definition of chain state. It was broken out specifically to reduce client overheads of versioning infrequently-changing nearly-primitive types like bigints. 

Versioning this library will be quite painful. Lotus and other clients of actors will need to adapt bigints wherever they come from chain state. @Stebalien would have a better idea that me of the overhead likely in doing so (it might not be that bad). But if you were going to do that, you could abstract the bigints _anyway_, without touching this library, into a local version with a more hand-holdy attitude toward uninitialised values.

> surface APIs are failing left and right due to missing payloads

I don't have a very good idea of problem details you're describing. If code is interacting with chain state, I think it needs to use this library, and be careful. Code that is _not_ interacting with chain state maybe shouldn't? I'm not sure if this is reasonable.

Kubuxu:
> Code that is not interacting with chain state maybe shouldn't? I'm not sure if this is reasonable.

Currently, it is a part of the exposed interface in Lotus and other APIs, for example, message value, gas params so on and defined using this library as currently, we use the same Message type for external and internal interfaces.

anorth:
Ok, so one position would be that Lotus and other APIs shouldn't be so tightly coupled to the actor state representation. In this instance, due to the very strict compatibility requirements, Lotus UX or implementation are not in a position to push changes on this library without (themselves) incurring large overhead due to versioning. You sorta have to take this as given, and decide if you want to re-use it as-is or not. If you decide not to, then you need to adapt to/from this bigint to interact with chain/state libraries. Since it's roughly the same work to either (a) adapt to a v2 of this library, or (b) adapt to a local bigint library independent of actors, I'd say (b) is better since Lotus would gain control and reduce coupling. I can of course see the unpalatable nature of having two very similar libraries for very similar needs; the justification would be the uniquely strict blockchain compatibility needs.

But let's zoom out a bit to a discussion of how to solve this problem, rather than how to make this solution work. As I understand it, another possible resolution is to change all callers of `IsZero` to use `IsNilOrZero`. I don't see anything standing in the way of this today. I'd be willing to mark `IsZero` deprecated to warn against future naive use, but we can't remove it. One might object to this as being imperfectly final as a fix, but I think it's quite pragmatic in light of, again, the uniquely strict blockchain compatibility needs.

Revising the problem statement further:

> These are often instantiated as part of a larger struct unmarshall

(and from elsewhere)

> There isn't much I can `... != nil` as the objects are there. They have just been instantiated incomplete by the marshaller

This sounds like the problem - your unmarshaller needs fixing! The actors and chain manipulation code more or less have no problem here because the CBOR de/serialization layer ensures these values are always instantiated non-nil. Your context is implicitly deserialising JSON, as I understand it. You want a policy of interpreting missing bigint values as zero. Can you fix whatever deserialization system is in use to enforce that, instead of changing the semantics of the library after nil initialisation.

PR #30, opened 2021-07-15
Add version 6.5

ZenGround0:
This is motivated by the effort to fix lotus pricelist to rely on version numbers https://github.com/filecoin-project/lotus/issues/6652

Turns out versions are already an enum so this was easy.


arajasek (approved):
noice

Stebalien:
Discussed in standup:

This should work, but it's exposed over the API so there are some concerns that users are using it. The _main_ use is, as far as I know, the lotus miner.

The _correct_ way to do this update is to make a breaking version change to the API, which is always a bit annoying.

ZenGround0:
@Stebalien how terrible would it be to change from int to float if we make a breaking api change?  The benefit would be that version numbers would look like users would expect: 6.5, 13.0.  The cost would be using floats.

Stebalien:
Awful? Checking equality between floats is generally a bad idea and not every (base-10) decimal can even be represented (accurately) as a float (which are effectively base-2 decimals).

If we're going to make a breaking change, I'd make these versions:

- 100
- 200
- 300
...

So we can stick a 250 in there.

We'd have to... rework a few cases where we assume that versions are mostly sequential, but that's not really a problem.

whyrusleeping:
Hey friends, people were using this and it not just breaks, it causes panics

whyrusleeping:
The panic we encountered comes from here: https://github.com/filecoin-project/lotus/blob/master/chain/actors/version.go#L37

Any version mismatch between this code and where youre getting the version number from will likely hit this.

Stebalien:
Ok, I _think_ we can sort of fix this by "aliasing" the old versions in that code. I.e., manually checking:

```go
if version < 100 {
    version = version*100
}
```

PR #32, opened 2021-10-11
Intoduce update proofs enums

Kubuxu:
The names include StackedDrg as the proof construction is specific to
PoRep.


abi/sector.go
```diff
@@ -223,6 +244,16 @@ func (p RegisteredSealProof) RegisteredWindowPoStProof() (RegisteredPoStProof, e
 	return info.WindowPoStProof, nil
 }
 
+// UpdateProof produces the PoSt-specific RegisteredProof corresponding
```
ZenGround0, line 247: s/PoSt/Update/g

PR #35, opened 2022-01-06
CBOR encode Addr Pairs for keys

arajasek:


abi/cbor_gen.go
```diff
@@ -163,3 +163,38 @@ func (t *SectorID) UnmarshalCBOR(r io.Reader) error {
 	}
 	return nil
 }
+
+var lengthBufAddrPairKey = []byte{128}
+
+func (t *AddrPairKey) MarshalCBOR(w io.Writer) error {
```
Stebalien, line 169: This is going to encode/decode nothing unless you make the struct members public.

Stebalien (changes requested):

abi/cbor_gen.go
arajasek: Oof, thank you, fixed

Stebalien (approved):

PR #36, opened 2022-04-22
Use new go-state-types state accessors

arajasek:
This rips out the state types from specs-actors, but not any of the actors logic. The goal is for this to be used by Go-based implementations of the Filecoin protocol that need convenient accessors for the state types and method parameters of the Filecoin builtin actors. 

No part of this is intended to be used for message execution.


builtin/network.go
```diff
@@ -0,0 +1,63 @@
...
+// The duration of a chain epoch.
+// Motivation: It guarantees that a block is propagated and WinningPoSt can be successfully done in time all supported miners.
+// Usage: It is used for deriving epoch-denominated periods that are more naturally expressed in clock time.
+// TODO: In lieu of a real configuration mechanism for this value, we'd like to make it a var so that implementations
+// can override it at runtime. Doing so requires changing all the static references to it in this repo to go through
+// late-binding function calls, or they'll see the "wrong" value.
+// https://github.com/filecoin-project/specs-actors/issues/353
+// If EpochDurationSeconds is changed, update `BaselineExponent`, `lambda`, and // `expLamSubOne` in ./reward/reward_logic.go
+// You can re-calculate these constants by changing the epoch duration in ./reward/reward_calc.py and running it.
+const EpochDurationSeconds = 30
```
vyzo, line 19: how will this work for devnets etc?
Maybe we should have build tags?

builtin/sector.go
```diff
@@ -0,0 +1,130 @@
...
+)
+
+// Policy values associated with a seal proof type.
+type SealProofPolicy struct {
+	SectorMaxLifetime stabi.ChainEpoch
+}
+
+// For V1 Stacked DRG sectors, the max is 540 days since Network Version 11
+// 	according to https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0014.md
+const EpochsIn540Days = stabi.ChainEpoch(540 * EpochsInDay)
```
vyzo, line 15: same problem exists here.

builtin/v8/manifest/manifest.go
```diff
@@ -0,0 +1,111 @@
+package manifest
```
vyzo, line 1: Maybe this should be outside the v8 structure?
We dont anticipate it changing, amd it is kind of self versioned.

builtin/v8/market/market_balances.go
```diff
@@ -0,0 +1,105 @@
+package market
+
+//func (m *marketStateMutation) lockClientAndProviderBalances(proposal *DealProposal) error {
```
vyzo, line 3: delete?

builtin/v8/market/policy.go
```diff
@@ -0,0 +1,69 @@
+package market
+
+import (
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/go-state-types/builtin"
+)
+
+// The number of epochs between payment and other state processing for deals.
+const DealUpdatesInterval = builtin.EpochsInDay // PARAM_SPEC
```
vyzo, line 10: the usual problem.

vyzo (commented):
Overall this looks good, and i am om board.

A couple of issues:
- the manifest type (and also, adt) should probably live outsied the v8 dir.
- bigger problem is the constants. Maybe we should make policy objects grouping them, similar to what we do in builtin-actors?

anorth:
Here is the venue for design discussion: https://github.com/filecoin-project/lotus/issues/8607

anorth (commented):
I think everything needs to go under the `v8` package, including `builtin`. Those are liable to change between versions too. Consumers like Lotus already had to deal with versioning of these.

builtin/network.go
anorth: Whatever was already working here should keep working. Let's not bring in new scope to this PR which should be essentially a mechanical move/rename.

builtin/v8/manifest/manifest.go
```diff
@@ -0,0 +1,111 @@
...
+	"io"
+
+	"github.com/filecoin-project/go-state-types/builtin/v8/util/adt"
+
+	"github.com/ipfs/go-cid"
+
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+type Manifest struct {
```
anorth, line 15: This has no documentation so I have no idea what it's for (probably already an issue before you moved it).

builtin/v8/market/market_params.go
```diff
@@ -0,0 +1,138 @@
...
+type PublishStorageDealsParams struct {
+	Deals []ClientDealProposal
+}
+
+type PublishStorageDealsReturn struct {
+	IDs        []abi.DealID
+	ValidDeals bitfield.BitField
+}
+
+// - Array of sectors rather than just one
```
anorth, line 28: It looks like you dropped the first line of this comment ("changes since version xxx")

builtin/v8/market/market_params.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+type OnMinerSectorsTerminateParams struct {
+	Epoch   abi.ChainEpoch
+	DealIDs []abi.DealID
+}
+
+// Validates a collection of deal dealProposals for activation, and returns their combined weight,
+// split into regular deal weight and verified deal weight.
+func ValidateDealsForActivation(
```
anorth, line 75: I can see why you retained this, but I'm not sure it belongs here. This is really just actor logic that happened to be the state file for some reason. While maintaining two language versions of the state schemas is reasonable, maintaining duplicate application logic is not.

This was exported for use by nodes, but in in the future a better approach to things like this will be to call a readonly/view method on an actor locally in order to use its logic. The market actor may even become non-builtin in the future, other market actors certainly won't be.

I don't want to block on this, but this approach won't be sustainable.

This belongs in state, rather than params, I think.

builtin/v8/market/market_state.go
```diff
@@ -0,0 +1,114 @@
...
+		TotalClientStorageFee:         abi.NewTokenAmount(0),
+	}, nil
+}
+
+////////////////////////////////////////////////////////////////////////////////
+// State utility functions
+////////////////////////////////////////////////////////////////////////////////
+
+// MarketStateMutationPermission is the mutation permission on a state field
+type MarketStateMutationPermission int
```
anorth, line 105: You shouldn't need this exported, it's an internal detail

builtin/v8/market/policy.go
anorth: These aren't really state, though I see why you wanted to export them similarly. Again, long term, we probably want the actors to encapsulate their logic/policy better, and access it via view methods.

builtin/v8/miner/bitfield_queue.go
```diff
@@ -0,0 +1,160 @@
...
+func LoadBitfieldQueue(store adt.Store, root cid.Cid, quant builtin.QuantSpec, bitwidth int) (BitfieldQueue, error) {
+	arr, err := adt.AsArray(store, root, bitwidth)
+	if err != nil {
+		return BitfieldQueue{}, xerrors.Errorf("failed to load epoch queue %v: %w", root, err)
+	}
+	return BitfieldQueue{arr, quant}, nil
+}
+
+// Adds values to the queue entry for an epoch.
+func (q BitfieldQueue) AddToQueue(rawEpoch abi.ChainEpoch, values bitfield.BitField) error {
```
anorth, line 32: These are an interesting case. There is logic here tightly coupled to the actor implementations. You've taken all of the mutation methods as well as inspection. It would be really annoying for future Go actor development if this logic were separate from the main actor logic like this – so it's a good think we're not doing any!

One possible justification for including this here is for future migrations (but the long term plan for that should be that they're written in Rust).

For now, I think I would suggest excluding all the mutation methods, and keeping this repo as thin as possible. Is there a case I'm missing that motivates including them?

builtin/v8/miner/deadline_assignment.go
```diff
@@ -0,0 +1,219 @@
...
+
+	"golang.org/x/xerrors"
+)
+
+// Helper types for deadline assignment.
+type deadlineAssignmentInfo struct {
+	index        int
+	liveSectors  uint64
+	totalSectors uint64
+}
```
anorth, line 14: This is logic, not state, it doesn't belong in this repo.

builtin/v8/miner/deadline_state.go
```diff
@@ -0,0 +1,1321 @@
...
+		}
+		err = cb(uint64(dlIdx), dl)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (d *Deadlines) UpdateDeadline(store adt.Store, dlIdx uint64, deadline *Deadline) error {
```
anorth, line 148: I don't think including all the mutation methods in this repo is appropriate. It's a very large amount of essentially application logic. There's no way  we'll keep it in sync with the canonical actor code.

builtin/v8/miner/expiration_queue.go
```diff
@@ -0,0 +1,881 @@
...
+		OnTimeSectors: onTimeSectors,
+		EarlySectors:  earlySectors,
+		OnTimePledge:  onTimePledge,
+		ActivePower:   activePower,
+		FaultyPower:   faultyPower,
+	}
+}
+
+// Adds sectors and power to the expiration set in place.
+func (es *ExpirationSet) Add(onTimeSectors, earlySectors bitfield.BitField, onTimePledge abi.TokenAmount, activePower, faultyPower PowerPair) error {
```
anorth, line 47: Application code that shoulnd't be here.

builtin/v8/miner/miner_state.go
```diff
@@ -0,0 +1,1225 @@
...
+type CollisionPolicy bool
+
+const (
+	DenyCollisions  = CollisionPolicy(false)
+	AllowCollisions = CollisionPolicy(true)
+)
+
+// Marks a set of sector numbers as having been allocated.
+// If policy is `DenyCollisions`, fails if the set intersects with the sector numbers already allocated.
+func (st *State) AllocateSectorNumbers(store adt.Store, sectorNos bitfield.BitField, policy CollisionPolicy) error {
```
anorth, line 313: Lots of application logic in this file.

builtin/v8/miner/monies.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"github.com/filecoin-project/go-state-types/builtin"
+	"github.com/filecoin-project/go-state-types/builtin/v8/util/math"
+	"github.com/filecoin-project/go-state-types/builtin/v8/util/smoothing"
+)
+
+// Projection period of expected sector block reward for deposit required to pre-commit a sector.
+// This deposit is lost if the pre-commitment is not timely followed up by a commitment proof.
+var PreCommitDepositFactor = 20 // PARAM_SPEC
```
anorth, line 13: This is mostly logic, not state

anorth (commented):
I'm down with this approach in general, but this PR is currently extracting way too much stuff. For the more complex actors, it's pulled out a huge amount of application logic that happens to be implemented on the state objects. This is actor code, mostly unnecessary for interpreting the state, and will be infeasible to keep in sync with the real version that counts, in builtin-actors. I didn't comment on all the occurrences of this, but there's a load in miner actor and others.


Since essentially everything in this repo represents a duplication of an idea expressed elsewhere, we should strive really hard to keep it as minimal as possible.

anorth (commented):
Reward testdata files don't belong here. Nor util/math, util/smoothing

builtin/network.go
ZenGround0: Strong agree with @anorth

arajasek:
From sync with @ZenGround0 : we should move the cbor generation into the root gen.go (and get it to generate cbor for all versions).

builtin/v8/market/market_params.go
arajasek: Yeah, you're quite that this is going to be unsustainable / will eventually diverge from the real source of truth (builtin actors). We'll open an issue for it.

Agreed that it should be in state, not params, though. Making that change now.

builtin/v8/miner/miner_types.go
```diff
@@ -0,0 +1,350 @@
...
+
+type AggregateSealVerifyProofAndInfos struct {
+	Miner          abi.ActorID
+	SealProof      abi.RegisteredSealProof
+	AggregateProof abi.RegisteredAggregationProof
+	Proof          []byte
+	Infos          []AggregateSealVerifyInfo
+}
+
+type ReplicaUpdateInfo struct {
```
ZenGround0, line 247: please move these guys out of v8

ZenGround0 (approved):
LGTM after moving proofs out of v8

ZenGround0:
LGTM

anorth (commented):
Definitely an improvement, thanks. 
- I think I understand why they're still there, but I'd love to see as much of the `policy.go` (and related `monies.go` etc) code removed if possible.
- Any chance of removing (part of) `reward_logic.go`?
- Ditto `util/smoothing/alpha_beta_filter.go`?

arajasek:
@anorth Unfortunately, I think everything left behind is used by Lotus today:

- The remaining `policy` and `monies` is needed for making decisions about deal collateral, how much to send with Pre/Prove commit messages, etc.
- `reward_logic` (what's left) is currently needed to construct the reward state, which is needed in order to be able to start new networks.
- `alpha_beta_filter` is used when performing calculations about `monies`.

PR #42, opened 2022-06-17
Deal label: Tweak JSON un/marshaling

arajasek:


magik6k (approved):
API breaking but lgtm; I don't see a better, less api breaking way to do this.

PR #44, opened 2022-06-30
fix: add missing FVM exitcode (4, 11 , and 22-24)

frrist:
Ported from:
- https://github.com/filecoin-project/ref-fvm/blob/17e5a31cee866da7045272d6f99a4b9a7d53a94e/shared/src/error/mod.rs#L64
- https://github.com/filecoin-project/ref-fvm/blob/17e5a31cee866da7045272d6f99a4b9a7d53a94e/shared/src/error/mod.rs#L77
- https://github.com/filecoin-project/ref-fvm/blob/17e5a31cee866da7045272d6f99a4b9a7d53a94e/shared/src/error/mod.rs#L100-L104

### DIFF
--- a/exitcode/common.go
+++ b/exitcode/common.go
@@ -4,18 +4,24 @@ package exitcode
 // Actors may also define their own codes, including redefining these values.
 
 const (
-	// Indicates a method parameter is invalid.
+	// ErrIllegalArgument Indicates a method parameter is invalid.
 	ErrIllegalArgument = FirstActorErrorCode + iota
-	// Indicates a requested resource does not exist.
+	// ErrNotFound Indicates a requested resource does not exist.
 	ErrNotFound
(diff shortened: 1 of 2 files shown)


arajasek:
Thanks for the PR! I wanna examine use cases of these exit codes in the clients, just to know how important they are. I think philosophically they belong here, though, and I'm sure there's uses for it (eg. Lily).

Stebalien:
Note: we now have a bunch of well-defined "user" exit codes as well. Exit codes are defined in https://github.com/filecoin-project/fvm-specs/blob/main/07-errors.md#exit-codes.

anorth (approved):

frrist:
@arajasek feel free to merge if this fits with the use cases you are looking into.

PR #46, opened 2022-08-04
Add a migration to go from v8 to v9

arajasek:
This pulls in the migration logic from the specs-actors repo. The migration only migrates the actor Code IDs and the system actor's state right now, which will be the starting point for all future migrations. Additional complexity will be introduced as more FIPs land in v9 (eg. FIP-0034).


jennijuju:
converting draft as it needs test

builtin/v9/migration/test/migration_test.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/filecoin-project/go-state-types/builtin/v9/migration"
+
+	"github.com/filecoin-project/go-state-types/builtin"
+	system9 "github.com/filecoin-project/go-state-types/builtin/v9/system"
+	"github.com/filecoin-project/go-state-types/builtin/v9/util/adt"
+	cbor "github.com/ipfs/go-ipld-cbor"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNv16Migration(t *testing.T) {
```
geoff-vball, line 20: This should be nv17

builtin/v9/migration/test/migration_test.go
```diff
@@ -0,0 +1,99 @@
...
+	require.NoError(t, err)
+
+	newSystemActor, found, err := newStateTree.GetActor(builtin.SystemActorAddr)
+	require.NoError(t, err)
+	require.True(t, found, "system actor not found")
+
+	var newSystemState system9.State
+	err = adtStore.Get(ctx, newSystemActor.Head, &newSystemState)
+	require.NoError(t, err)
+	require.Equal(t, newManifestDataCid, newSystemState.BuiltinActors)
```
geoff-vball, line 73: We should leave a comment on this test to change this to check the states of the actors that we're changing in this upgrade (or maybe just set something up to check them all).

builtin/v9/migration/test/migration_test.go
```diff
@@ -0,0 +1,99 @@
...
+	require.NoError(t, err)
+
+	cacheRoot, err := migration.MigrateStateTree(ctx, adtStore, newManifestCid, startRoot, 200, migration.Config{MaxWorkers: 1}, log, cache)
+	require.NoError(t, err)
+
+	noCacheRoot, err := migration.MigrateStateTree(ctx, adtStore, newManifestCid, startRoot, 200, migration.Config{MaxWorkers: 1}, log, migration.NewMemMigrationCache())
+	require.NoError(t, err)
+	require.True(t, cacheRoot.Equals(noCacheRoot))
+
+	// TODO: check all the CIDs
```
geoff-vball, line 59: Get rid of this TODO (in both versions of this test). We compare the CIDs on line 93

builtin/v9/migration/test/util.go
```diff
@@ -0,0 +1,147 @@
...
+
+	manifestCid8, manifestDataCid8 := makeTestManifest(t, store, "fil/8/")
+	var manifest8 manifest.Manifest
+	require.NoError(t, store.Get(ctx, manifestCid8, &manifest8), "error reading actor manifest")
+	require.NoError(t, manifest8.Load(ctx, store), "error loading actor manifest")
+
+	accountCid, ok := manifest8.Get("account")
+	require.True(t, ok, "didn't find account actor in manifest")
+
+	systemCid, ok := manifest8.Get("system")
```
geoff-vball, line 76: Is this all supposed to be v8 or v9 stuff?

builtin/v9/migration/top.go
```diff
@@ -0,0 +1,341 @@
...
+		}
+
+		migrations[code7Cid] = codeMigrator{code8Cid}
+	}
+
+	// migrations that migrate both code and state
+	system8Cid, ok := newManifest.Get("system")
+	if !ok {
+		return cid.Undef, xerrors.Errorf("code cid for system actor not found in manifest")
+	}
```
geoff-vball, line 141: These and other things in this file should be updated for the new versions. If we expect to duplicate this code again in the future, it might be best to change names to `old` and `new`.

builtin/v9/migration/test/migration_test.go
arajasek: Good call, I'll generalize the names

builtin/v9/migration/test/migration_test.go
arajasek: Yessir

builtin/v9/migration/test/migration_test.go
arajasek: What do you mean by "in both versions of this test"?

geoff-vball: This one and the one you copied it from.

builtin/v9/migration/top.go
arajasek: yessir

geoff-vball (approved):

PR #48, opened 2022-08-04
feat: migration for FIP-0034

arajasek:
The specification for the change can be found [here](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0034.md), and the Rust-based implementation can be found [here](https://github.com/filecoin-project/builtin-actors/pull/484).


anorth (commented):
Thanks for this!

jennijuju:
https://github.com/filecoin-project/go-commp-utils/pull/9

builtin/v9/migration/top.go
```diff
@@ -140,8 +150,35 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
+	if !ok {
+		return cid.Undef, xerrors.New("didn't find market actor")
+	}
+
+	var marketState market8.State
+	if err := store.Get(ctx, marketActor.Head, &marketState); err != nil {
+		return cid.Undef, xerrors.Errorf("failed to get system actor state: %w", err)
+	}
+
+	proposals, err := market8.AsDealProposalArray(adtStore, marketState.Proposals)
```
arajasek, line 175: I need someone to talk me through thread-safety here. I'm pretty sure this _isn't_ safe. 

If so, do we need to load the proposals AMT each time we migrate a miner [with deals in their pending precommits]? That wouldn't be too bad (ultimately not _that_ many such loads), but I'm wondering if people have better ideas.

builtin/v9/miner/miner_types.go
```diff
@@ -53,6 +53,21 @@ type ProveReplicaUpdatesParams struct {
...
+	Partition            uint64
+	NewSealedSectorCID   cid.Cid `checked:"true"`
+	NewUnsealedSectorCID cid.Cid `checked:"true"`
+	Deals                []abi.DealID
+	UpdateProofType      abi.RegisteredUpdateProof
+	ReplicaProof         []byte
+}
+
+type ProveReplicaUpdatesParams2 struct {
+	Updates []ReplicaUpdate
```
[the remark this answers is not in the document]
arajasek: Fixed, thanks for the catch

builtin/v9/migration/miner.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+
+	newHead, err := store.Put(ctx, &outState)
+	return &actorMigrationResult{
+		newCodeCID: m.migratedCodeCID(),
+		newHead:    newHead,
+	}, err
+}
+
+func generateUnsealedCID(proofType abi.RegisteredSealProof, pieces []abi.PieceInfo) (cid.Cid, error) {
```
arajasek, line 133: This method became a lot smaller after I took out the padding logic (which the commp util already does). I'll flatten it into the caller.

builtin/v9/migration/miner.go
```diff
@@ -0,0 +1,141 @@
...
+
+		err = newPrecommitOnChainInfos.Put(miner9.SectorKey(info.Info.SectorNumber), &miner9.SectorPreCommitOnChainInfo{
+			Info: miner9.SectorPreCommitInfo{
+				SealProof:     info.Info.SealProof,
+				SectorNumber:  info.Info.SectorNumber,
+				SealedCID:     info.Info.SealedCID,
+				SealRandEpoch: info.Info.SealRandEpoch,
+				DealIDs:       info.Info.DealIDs,
+				Expiration:    info.Info.Expiration,
+				UnsealedCid:   unsealedCid,
```
ribasushi, line 86: All looks good. Idly wondering why is this is a `*cid.Cid` as opposed to `cid.Cid`. Is it to allow for the optional/assumed CC CommD?

builtin/v9/migration/miner.go
ribasushi: They are like 1:1 identical signatures now no?

geoff-vball (approved):

builtin/v9/migration/miner.go
arajasek: Yes they are.

PR #62, opened 2022-09-02
Add miner state to v9 needed for FIP-0029

geoff-vball:
These state changes should match up with https://github.com/filecoin-project/builtin-actors/pull/496/files

### DIFF
--- a/builtin/v9/miner/miner_state.go
+++ b/builtin/v9/miner/miner_state.go
@@ -94,12 +94,12 @@ type MinerInfo struct {
 	// This is the address that tokens will be withdrawn to
 	Beneficiary addr.Address
 
-	// Beneficiary's withdrawl quota, how much of the quota has been withdrawn,
+	// Beneficiary's withdrawal quota, how much of the quota has been withdrawn,
 	// and when the Beneficiary expires.
 	BeneficiaryTerm BeneficiaryTerm
 
-	// A proposal new beneficiary message for this miner
-	PendingBeneficiaryTerm *BeneficiaryTermChange
+	// A proposed change to `BenificiaryTerm`
+	PendingBeneficiaryTerm *BeneficiaryChange
 
 	// Additional addresses that are permitted to submit messages controlling this actor (optional).
 	ControlAddresses []addr.Address // Must all be ID addresses.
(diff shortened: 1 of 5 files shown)


builtin/v9/miner/miner_state.go
```diff
@@ -90,6 +90,17 @@ type MinerInfo struct {
 	// The associated pubkey-type address is used to sign blocks and messages on behalf of this miner.
 	Worker addr.Address // Must be an ID-address.
 
+	// Beneficiary address for this miner.
+	// This is the address that tokens will be withdrawn to
+	Beneficiary addr.Address
+
+	// Beneficiary's withdrawl quota, how much of the quota has been withdrawn,
```
arajasek, line 97:
```suggestion
	// Beneficiary's withdrawa quota, how much of the quota has been withdrawn,
```

builtin/v9/miner/miner_state.go
```diff
@@ -90,6 +90,17 @@ type MinerInfo struct {
...
+	// and when the Beneficiary expires.
+	BeneficiaryTerm BeneficiaryTerm
+
+	// A proposal new beneficiary message for this miner
```
arajasek, line 101:
```suggestion
	// A proposed change to `BenificiaryTerm`
```

builtin/v9/miner/miner_state.go
```diff
@@ -178,6 +189,20 @@ func (st *State) GetInfo(store adt.Store) (*MinerInfo, error) {
 	return &info, nil
 }
 
+type BeneficiaryTerm struct {
+	Quota      abi.TokenAmount
+	UsedQuota  abi.TokenAmount
+	Expiration abi.ChainEpoch
+}
+
+type BeneficiaryTermChange struct {
```
arajasek, line 198: I'd either match the name in builtin actors, or PR a change to the name there (I think this name is better tbh).

arajasek (commented):
A few things missing from this:

- the params and return of the 2 new methods need to be added
- all new de/serializable types need to be added to `v9/gen/gen.go`
- cbor gen needs to be run

builtin/v9/miner/miner_state.go
```diff
@@ -178,6 +189,20 @@ func (st *State) GetInfo(store adt.Store) (*MinerInfo, error) {
 	return &info, nil
 }
 
+type BeneficiaryTerm struct {
+	Quota      abi.TokenAmount
+	UsedQuota  abi.TokenAmount
+	Expiration abi.ChainEpoch
+}
+
+type BeneficiaryChange struct {
```
arajasek, line 198: I'd either match the name in builtin actors, or PR a change to the name there (I think this name is better tbh).

arajasek (commented):
One last nit, but otherwise you're G2G.

arajasek (approved):

Issue #55, opened 2022-08-26
FIP 0029 miner state migration


jennijuju:
state changes - adding three states to miner info state  https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0029.md#specification

the beneficiary should be set as owner address upon migration

PR #65, opened 2022-09-06
feat: migration for FIP-0029

geoff-vball:
Closes https://github.com/filecoin-project/go-state-types/issues/55

The specification for the change can be found [here](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0029.md), and the Rust-based implementation can be found [here](https://github.com/filecoin-project/builtin-actors/pull/496/files).


arajasek (commented):
Looks good, needs test

arajasek (approved):

PR #70, opened 2022-09-16
feat: actors: Implement datacap actor

geoff-vball:
Mirrors datacap actor implementation found in dev branch https://github.com/filecoin-project/builtin-actors/tree/decouple-fil%2B


arajasek (commented):
Can we integrate this into Lotus in a draft PR too? It looks mostly correct, but I suspect there might be dragons that we might as well slay before merging.

arajasek (approved):

PR #79, opened 2022-09-23
feat: crypto: add SigType: SigTypeDelegated

ychiaoli18:
A new SigType is required for sig.Verify to check the signature signed by Ethereum's clients. The reason we can't use existing SigTypeSecp256k1 is that its verifier uses blake2b to hash the message, but SigTypeDelegated's verifier should use keccak256 to hash the message, so we need a new SigType to distinguish them.

### DIFF
--- a/crypto/signature.go
+++ b/crypto/signature.go
@@ -16,7 +16,7 @@ const (
 
 	SigTypeSecp256k1 = SigType(iota)
 	SigTypeBLS
-	SigTypeEthSecp256k1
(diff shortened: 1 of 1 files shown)


raulk (commented):
Could you please call this `Delegated`? That's what the account abstraction delegated signature type is shaping up to be. That way we won't merge something we know we're gonna have to back out.

raulk (approved):

PR #81, opened 2022-09-23
feat: crypto: add SigType: SigTypeDelegated

ychiaoli18:
This should supersede https://github.com/filecoin-project/go-state-types/pull/79 since the target branch should be experimental/fvm-m2. 

A new SigType is required for sig.Verify to check the signature signed by Ethereum's clients. The reason we can't use existing SigTypeSecp256k1 is that its verifier uses blake2b to hash the message, but SigTypeDelegated's verifier should use keccak256 to hash the message, so we need a new SigType to distinguish them.

### DIFF
--- a/crypto/signature.go
+++ b/crypto/signature.go
@@ -16,7 +16,7 @@ const (
 
 	SigTypeSecp256k1 = SigType(iota)
 	SigTypeBLS
-	SigTypeEthSecp256k1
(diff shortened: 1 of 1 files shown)


arajasek:
See https://github.com/filecoin-project/go-state-types/pull/82#issuecomment-1257012081

PR #82, opened 2022-09-23
Revert "feat: crypto: add SigType: SigTypeDelegated"

ychiaoli18:
Reverts filecoin-project/go-state-types#79


arajasek:
@ychiaoli18 I think it's fine to leave this in `master` (and rebase `experimental/fvm-m2` onto `master`). Should have no impact until the new sig type actually starts showing up.

raulk:
It's ok for the original change to remain on master -- no need to revert!

Issue #69, opened 2022-09-16
FIP-0045: Migration


arajasek:
Here is the proposed  workflow for the FIP-0045 migration:

- [ ] Step 0: Migrate the MinerActors (happens as part of the broader migration)
  - SectorOnChainInfos are updated in the Sectors AMT as well as SectorsSnapshot AMTs in each Deadline
- [ ] Step 1: Create the DataCap actor, and put it into the state tree
  - This only needs the v8 VerifiedRegistry's VerifiedClients field as input, with tokens being granted to each client
-  [ ] Step 2: Migrate the VerifregActor
   - The input needed is a `map[DealID]DealProposal` of verified deals in the PendingProposals Set
   - This can be constructed from the v8 market state, and provided as input
   - For each of these deals, create an Allocation in the Allocations HAMT
   - Return a `map[DealID]AllocationID` for input in the next step

- [ ] Step 3: Migrate the MarketActor 
  - This requires the `map` result of the previous step (the Verifreg migration) as input
  - This map effectively becomes the `PendingDealAllocationIds`

arajasek:
Further caching speedups:

- cache precommits for the miner ('migratePrecommits`)

arajasek:
Open question: https://github.com/filecoin-project/go-state-types/pull/85#discussion_r989550336

PR #83, opened 2022-09-30
FIP-0045: Miner migration

arajasek:
Step 0 of #69

The minerMigrator performs the following migrations:
- FIP-0029: Sets the Beneficary to the Owner, and sets empty values for BeneficiaryTerm and PendingBeneficiaryTerm
- FIP-0034: For each SectorPreCommitOnChainInfo in PreCommitedSectors, calculates the unsealed CID (assuming there are deals)
- FIP-0045: For each SectorOnChainInfo in Sectors, set SimpleQAPower = (DealWeight == 0 && VerifiedDealWeight == 0)
- FIP-0045: For each Deadline in Deadlines: for each SectorOnChainInfo in SectorsSnapshot, set SimpleQAPower = (DealWeight == 0 && VerifiedDealWeight == 0)

The migration has the following layers of caching:

- at the `ActorHead` level -- if a matching (address, Head) tuple is found in the cache, use the cached value
- at the `SectorsAmt` level -- if a matching Sectors CID is found in the cache, use the cached value
  - This is used for migrating both the Sectors AMT, as well as the SectorsSnapshot AMTs in each deadline
- at the "previous SectorsAmt" level -- if we have the previous input and output of migrating this particular miner's SectorsAmt before, load them, and transform previous output to current output based on the diff between previous input and current input

Further caching is possible at the following levels;
- at the PreCommits level -- if a matching PreCommittedSectors CID is found in the cache, use the cached value
- at the individual PreCommit level -- if a matching SectorPreCommitOnChainInfo CID is found in the cache, use the cached value

### DIFF
--- a/builtin/v8/miner/deadline_state.go
+++ b/builtin/v8/miner/deadline_state.go
@@ -96,8 +96,55 @@ const DeadlineExpirationAmtBitwidth = 5
 const DeadlineOptimisticPoStSubmissionsAmtBitwidth = 2
 
 //
+// Deadline (singular)
+//
+
+func ConstructDeadline(store adt.Store) (*Deadline, error) {
+	emptyPartitionsArrayCid, err := adt.StoreEmptyArray(store, DeadlinePartitionsAmtBitwidth)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty partitions array: %w", err)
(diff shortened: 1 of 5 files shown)


arajasek:
Lumped into #85

PR #84, opened 2022-10-01
FIP-0045: DataCap migration

arajasek:
Step #1 of #69

The DataCap actor is created, and its State is set based on the Verified Registry's VerifiedClients map. For each client:
- its balance is converted to tokens that are added to the TokenState
- an entry is created in the TokenState's `allowances` map, with infinite allowance and the Market actor as operator


arajasek:
Lumped into #85

PR #85, opened 2022-10-02
FIP-0045: Implement migrations

### LINKED ISSUE #69: FIP-0045: Migration

arajasek:
Closes #69

**The sequence of steps is:**

- Step 0: Add the DataCap actor to the v8 state tree with a fake Head
- Step 1A: Migrate all actors except the StorageMarketActor and VerifiedRegistryActor as part of the "regular" migration workflow (more details on the migrations below), set the results in the output state tree
- Step 1B: In a separate, parallel goroutine migrate the VerifiedRegistryActor, and then the StorageMarketActor
- Block on 1A and 1B
- Step 2: Set the VerifiedRegistryActor and StorageMarketActor in the output state tree
- Step 3: Flush and return

**Miner Migration**

The minerMigrator performs the following migrations:
- FIP-0029: Sets the `Beneficary` to the `Owner`, and sets empty values for `BeneficiaryTerm and PendingBeneficiaryTerm`
- FIP-0034: For each `SectorPreCommitOnChainInfo` in `PreCommitedSectors`, calculates the unsealed CID (assuming there are deals)
- FIP-0045: For each SectorOnChainInfo in Sectors, set `SimpleQAPower = (DealWeight == 0 && VerifiedDealWeight == 0)`
- FIP-0045: For each Deadline in Deadlines: for each `SectorOnChainInfo` in `SectorsSnapshot`, set `SimpleQAPower = (DealWeight == 0 && VerifiedDealWeight == 0)`

The migration has the following layers of caching:

- at the `ActorHead` level -- if a matching `(address, Head)` tuple is found in the cache, use the cached value
- at the `SectorsAmt` level -- if a matching `Sectors` CID is found in the cache, use the cached value
  - This is used for migrating both the `Sectors` AMT, as well as the `SectorsSnapshot` AMTs in each deadline
- at the "previous SectorsAmt" level -- if we have the previous input and output of migrating this particular miner's SectorsAmt before, load them, and transform previous output to current output based on the diff between previous input and current input

Further caching is possible at the following levels;
- at the `PreCommits` level -- if a matching `PreCommittedSectors` CID is found in the cache, use the cached value
- at the individual `PreCommit` level -- if a matching `SectorPreCommitOnChainInfo` CID is found in the cache, use the cached value

**DataCap "migration"**

The `DataCap` State is set based on the Verified Registry's `VerifiedClients` map. For each client:
- its balance is converted to tokens that are added to the `TokenState`
- an entry is created in the TokenState's `allowances` map, with infinite allowance and the Market actor as operator

**VerifiedRegistry & Market migration**

As in #69

### DIFF
--- a/builtin/shared.go
+++ b/builtin/shared.go
@@ -10,6 +10,8 @@ import (
 // This value has been empirically chosen, but the optimal value for maps with different mutation profiles may differ.
 const DefaultHamtBitwidth = 5
 
+const DefaultTokenActorBitwidth = 3
+
 type BigFrac struct {
 	Numerator   big.Int
 	Denominator big.Int
--- a/builtin/v8/util/adt/set.go
+++ b/builtin/v8/util/adt/set.go
@@ -1,0 +1,71 @@
+package adt
+
+import (
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/ipfs/go-cid"
+)
+
+// Set interprets a Map as a set, storing keys (with empty values) in a HAMT.
+type Set struct {
+	m *Map
+}
+
+// AsSet interprets a store as a HAMT-based set with root `r`.
+// The HAMT is interpreted with branching factor 2^bitwidth.
+func AsSet(s Store, r cid.Cid, bitwidth int) (*Set, error) {
+	m, err := AsMap(s, r, bitwidth)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Set{
+		m: m,
+	}, nil
+}
+
+// NewSet creates a new HAMT with root `r` and store `s`.
+// The HAMT has branching factor 2^bitwidth.
+func MakeEmptySet(s Store, bitwidth int) (*Set, error) {
+	m, err := MakeEmptyMap(s, bitwidth)
+	if err != nil {
+		return nil, err
+	}
+	return &Set{m}, nil
+}
+
+// Root return the root cid of HAMT.
+func (h *Set) Root() (cid.Cid, error) {
+	return h.m.Root()
+}
+
+// Put adds `k` to the set.
+func (h *Set) Put(k abi.Keyer) error {
+	return h.m.Put(k, nil)
+}
+
+// Has returns true iff `k` is in the set.
+func (h *Set) Has(k abi.Keyer) (bool, error) {
+	return h.m.Get(k, nil)
+}
+
+// Removes `k` from the set, if present.
+// Returns whether the key was previously present.
+func (h *Set) TryDelete(k abi.Keyer) (bool, error) {
+	return h.m.TryDelete(k)
+}
+
+// Removes `k` from the set, expecting it to be present.
+func (h *Set) Delete(k abi.Keyer) error {
+	return h.m.Delete(k)
+}
+
+// ForEach iterates over all values in the set, calling the callback for each value.
+// Returning error from the callback stops the iteration.
+func (h *Set) ForEach(cb func(k string) error) error {
+	return h.m.ForEach(nil, cb)
+}
+
+// Collects all the keys from the set into a slice of strings.
+func (h *Set) CollectKeys() (out []string, err error) {
+	return h.m.CollectKeys()
+}
--- a/builtin/v9/migration/miner.go
+++ b/builtin/v9/migration/miner.go
@@ -3,13 +3,17 @@ package migration
 import (
 	"context"
 
+	adt9 "github.com/filecoin-project/go-state-types/builtin/v9/util/adt"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-amt-ipld/v4"
 	"golang.org/x/xerrors"
 
 	commp "github.com/filecoin-project/go-commp-utils/nonffi"
 	"github.com/filecoin-project/go-state-types/builtin"
 	"github.com/filecoin-project/go-state-types/builtin/v8/market"
 	miner8 "github.com/filecoin-project/go-state-types/builtin/v8/miner"
-	"github.com/filecoin-project/go-state-types/builtin/v8/util/adt"
+	adt8 "github.com/filecoin-project/go-state-types/builtin/v8/util/adt"
 	miner9 "github.com/filecoin-project/go-state-types/builtin/v9/miner"
 
 	"github.com/ipfs/go-cid"
@@ -18,9 +22,72 @@ import (
 	"github.com/filecoin-project/go-state-types/abi"
 )
 
+// The minerMigrator performs the following migrations:
+// FIP-0029: Sets the Beneficary to the Owner, and sets empty values for BeneficiaryTerm and PendingBeneficiaryTerm
+// FIP-0034: For each SectorPreCommitOnChainInfo in PreCommitedSectors, calculates the unsealed CID (assuming there are deals)
+// FIP-0045: For each SectorOnChainInfo in Sectors, set SimpleQAPower = (DealWeight == 0 && VerifiedDealWeight == 0)
+// FIP-0045: For each Deadline in Deadlines: for each SectorOnChainInfo in SectorsSnapshot, set SimpleQAPower = (DealWeight == 0 && VerifiedDealWeight == 0)
+
 type minerMigrator struct {
-	proposals  *market.DealArray
-	OutCodeCID cid.Cid
+	emptyPrecommitOnChainInfosV9 cid.Cid
+	emptyDeadlineV8              cid.Cid
+	emptyDeadlinesV8             cid.Cid
+	emptyDeadlineV9              cid.Cid
+	emptyDeadlinesV9             cid.Cid
+	proposals                    *market.DealArray
+	OutCodeCID                   cid.Cid
+}
+
+func newMinerMigrator(ctx context.Context, store cbor.IpldStore, marketProposals *market.DealArray, outCode cid.Cid) (*minerMigrator, error) {
+	ctxStore := adt8.WrapStore(ctx, store)
+
+	emptyPrecommitMapCidV9, err := adt9.StoreEmptyMap(ctxStore, builtin.DefaultHamtBitwidth)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty precommit map v9: %w", err)
+	}
+
+	edv8, err := miner8.ConstructDeadline(ctxStore)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadline v8: %w", err)
+	}
+
+	edv8cid, err := store.Put(ctx, edv8)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to put empty deadline v8: %w", err)
+	}
+
+	edsv8 := miner8.ConstructDeadlines(edv8cid)
+	edsv8cid, err := store.Put(ctx, edsv8)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadlines v8: %w", err)
+	}
+
+	edv9, err := miner9.ConstructDeadline(ctxStore)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadline v9: %w", err)
+	}
+
+	edv9cid, err := store.Put(ctx, edv9)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to put empty deadline v9: %w", err)
+	}
+
+	edsv9 := miner9.ConstructDeadlines(edv9cid)
+	edsv9cid, err := store.Put(ctx, edsv9)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadlines v9: %w", err)
+
+	}
+
+	return &minerMigrator{
+		emptyPrecommitOnChainInfosV9: emptyPrecommitMapCidV9,
+		emptyDeadlineV8:              edv8cid,
+		emptyDeadlinesV8:             edsv8cid,
+		emptyDeadlineV9:              edv9cid,
+		emptyDeadlinesV9:             edsv9cid,
+		proposals:                    marketProposals,
+		OutCodeCID:                   outCode,
+	}, nil
 }
 
 func (m minerMigrator) migratedCodeCID() cid.Cid {
@@ -36,21 +103,90 @@ func (m minerMigrator) migrateState(ctx context.Context, store cbor.IpldStore, i
 	if err := store.Get(ctx, inState.Info, &inInfo); err != nil {
 		return nil, err
 	}
-	wrappedStore := adt.WrapStore(ctx, store)
+	wrappedStore := adt8.WrapStore(ctx, store)
 
-	oldPrecommitOnChainInfos, err := adt.AsMap(wrappedStore, inState.PreCommittedSectors, builtin.DefaultHamtBitwidth)
+	newPrecommits, err := m.migratePrecommits(ctx, wrappedStore, inState.PreCommittedSectors)
 	if err != nil {
-		return nil, xerrors.Errorf("failed to load old precommit onchain infos for miner %s: %w", in.address, err)
+		return nil, xerrors.Errorf("failed to migrate precommits for miner: %s: %w", in.address, err)
 	}
 
-	emptyMap, err := adt.StoreEmptyMap(wrappedStore, builtin.DefaultHamtBitwidth)
+	newSectors, err := migrateSectorsWithCache(ctx, wrappedStore, in.cache, in.address, inState.Sectors)
 	if err != nil {
-		return nil, xerrors.Errorf("failed to make empty map: %w", err)
+		return nil, xerrors.Errorf("failed to migrate sectors for miner: %s: %w", in.address, err)
 	}
 
-	newPrecommitOnChainInfos, err := adt.AsMap(wrappedStore, emptyMap, builtin.DefaultHamtBitwidth)
+	deadlinesOut, err := m.migrateDeadlines(ctx, wrappedStore, in.cache, inState.Deadlines)
 	if err != nil {
-		return nil, xerrors.Errorf("failed to load empty map: %w", err)
+		return nil, xerrors.Errorf("failed to migrate deadlines: %w", err)
+	}
+
+	var newPendingWorkerKey *miner9.WorkerKeyChange
+	if inInfo.PendingWorkerKey != nil {
+		newPendingWorkerKey = &miner9.WorkerKeyChange{
+			NewWorker:   inInfo.PendingWorkerKey.NewWorker,
+			EffectiveAt: inInfo.PendingWorkerKey.EffectiveAt,
+		}
+	}
+
+	outInfo := miner9.MinerInfo{
+		Owner:       inInfo.Owner,
+		Worker:      inInfo.Worker,
+		Beneficiary: inInfo.Owner,
+		BeneficiaryTerm: miner9.BeneficiaryTerm{
+			Quota:      abi.NewTokenAmount(0),
+			UsedQuota:  abi.NewTokenAmount(0),
+			Expiration: 0,
+		},
+		PendingBeneficiaryTerm:     nil,
+		ControlAddresses:           inInfo.ControlAddresses,
+		PendingWorkerKey:           newPendingWorkerKey,
+		PeerId:                     inInfo.PeerId,
+		Multiaddrs:                 inInfo.Multiaddrs,
+		WindowPoStProofType:        inInfo.WindowPoStProofType,
+		SectorSize:                 inInfo.SectorSize,
+		WindowPoStPartitionSectors: inInfo.WindowPoStPartitionSectors,
+		ConsensusFaultElapsed:      inInfo.ConsensusFaultElapsed,
+		PendingOwnerAddress:        inInfo.PendingOwnerAddress,
+	}
+	newInfoCid, err := store.Put(ctx, &outInfo)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to flush new miner info: %w", err)
+	}
+
+	outState := miner9.State{
+		Info:                       newInfoCid,
+		PreCommitDeposits:          inState.PreCommitDeposits,
+		LockedFunds:                inState.LockedFunds,
+		VestingFunds:               inState.VestingFunds,
+		FeeDebt:                    inState.FeeDebt,
+		InitialPledge:              inState.InitialPledge,
+		PreCommittedSectors:        newPrecommits,
+		PreCommittedSectorsCleanUp: inState.PreCommittedSectorsCleanUp,
+		AllocatedSectors:           inState.AllocatedSectors,
+		Sectors:                    newSectors,
+		ProvingPeriodStart:         inState.ProvingPeriodStart,
+		CurrentDeadline:            inState.CurrentDeadline,
+		Deadlines:                  deadlinesOut,
+		EarlyTerminations:          inState.EarlyTerminations,
+		DeadlineCronActive:         inState.DeadlineCronActive,
+	}
+
+	newHead, err := store.Put(ctx, &outState)
+	return &actorMigrationResult{
+		newCodeCID: m.migratedCodeCID(),
+		newHead:    newHead,
+	}, err
+}
+
+func (m minerMigrator) migratePrecommits(ctx context.Context, wrappedStore adt8.Store, inRoot cid.Cid) (cid.Cid, error) {
+	oldPrecommitOnChainInfos, err := adt8.AsMap(wrappedStore, inRoot, builtin.DefaultHamtBitwidth)
+	if err != nil {
+		return cid.Undef, xerrors.Errorf("failed to load old precommit onchain infos: %w", err)
+	}
+
+	newPrecommitOnChainInfos, err := adt9.AsMap(wrappedStore, m.emptyPrecommitOnChainInfosV9, builtin.DefaultHamtBitwidth)
+	if err != nil {
+		return cid.Undef, xerrors.Errorf("failed to load empty map: %w", err)
 	}
 
 	var info miner8.SectorPreCommitOnChainInfo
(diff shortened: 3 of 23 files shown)


builtin/v8/util/adt/set.go
```diff
@@ -0,0 +1,71 @@
+package adt
+
+import (
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/ipfs/go-cid"
+)
+
+// Set interprets a Map as a set, storing keys (with empty values) in a HAMT.
```
ZenGround0, line 8: nit: Since this is code and not just structs it would be nice to see a link to where this is copied from (if it is copied)

builtin/v9/migration/top.go
```diff
@@ -307,7 +350,317 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
 
 	elapsed := time.Since(startTime)
 	rate := float64(doneCount) / elapsed.Seconds()
-	log.Log(rt.INFO, "All %d done after %v (%.0f/s). Flushing state tree root.", doneCount, elapsed, rate)
+	log.Log(rt.INFO, "All %d done after %v (%.0f/s). Starting deferred migrations.", doneCount, elapsed, rate)
+
+	// Create the Datacap actor
```
ZenGround0, line 355: For clarity if possible please split into three different migration functions with inputs and outputs.  From talking offline it sounds like this should be possible.

builtin/v9/migration/miner.go
```diff
@@ -18,94 +24,102 @@ import (
...
+	}
+	var inInfo miner8.MinerInfo
+	if err := store.Get(ctx, inState.Info, &inInfo); err != nil {
+		return nil, err
+	}
+	wrappedStore := adt8.WrapStore(ctx, store)
 
-		return nil
-	})
+	newPrecommits, err := m.migratePrecommits(ctx, wrappedStore, inState.PreCommittedSectors)
```
ZenGround0, line 110: Potentially worth caching

builtin/v9/migration/miner.go
```diff
@@ -151,10 +165,10 @@ func (m minerMigrator) migrateState(ctx context.Context, store cbor.IpldStore, i
 		PreCommittedSectors:        newPrecommits,
 		PreCommittedSectorsCleanUp: inState.PreCommittedSectorsCleanUp,
 		AllocatedSectors:           inState.AllocatedSectors,
-		Sectors:                    inState.Sectors,
+		Sectors:                    newSectors,
```
ZenGround0, line 168: nit: lets use consistent naming between newSectors and deadlinesOut

builtin/v9/migration/miner.go
```diff
@@ -165,3 +179,262 @@ func (m minerMigrator) migrateState(ctx context.Context, store cbor.IpldStore, i
...
+	}
+
+	var info miner8.SectorPreCommitOnChainInfo
+	err = oldPrecommitOnChainInfos.ForEach(&info, func(key string) error {
+		var unsealedCid *cid.Cid
+		var pieces []abi.PieceInfo
+		for _, dealID := range info.Info.DealIDs {
+			deal, err := m.proposals.GetDealProposal(dealID)
+			if err != nil {
+				// Possible for the proposal to be missing if it's expired (but the deal is still in a precommit that's yet to be cleaned up)
```
ZenGround0, line 201: Nice let's make sure we test this

builtin/v9/migration/miner.go
```diff
@@ -165,3 +179,262 @@ func (m minerMigrator) migrateState(ctx context.Context, store cbor.IpldStore, i
...
+	if err = inArray.ForEach(&sectorInfo, func(k int64) error {
+		return outArray.Set(uint64(k), migrateSectorInfo(sectorInfo))
+	}); err != nil {
+		return nil, err
+	}
+
+	return outArray, err
+}
+
+// Need to introduce caching here too
```
ZenGround0, line 356: Is this now addressed with the cache.Load()?

builtin/v9/migration/miner.go
```diff
@@ -165,3 +179,262 @@ func (m minerMigrator) migrateState(ctx context.Context, store cbor.IpldStore, i
...
+			}
+
+		}
+
+		outRoot, err := outArray.Root()
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("error writing new sectors AMT: %w", err)
+		}
+
+		_ = cache.Write(MinerPrevSectorsInKey(minerAddr), inRoot)
```
ZenGround0, line 333: We probably don't want to swallow these errors.

builtin/v9/migration/miner.go
```diff
@@ -165,3 +179,262 @@ func (m minerMigrator) migrateState(ctx context.Context, store cbor.IpldStore, i
...
+
+		okOut, prevOutRoot, err := cache.Read(MinerPrevSectorsOutKey(minerAddr))
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("failed to get previous outRoot from cache: %w", err)
+		}
+
+		var outArray *adt9.Array
+		if okIn && okOut {
+			// we have previous work, but the AMT has changed -- diff them
+			diffs, err := amt.Diff(ctx, store, store, prevInRoot, inRoot, amt.UseTreeBitWidth(miner9.SectorsAmtBitwidth))
```
ZenGround0, line 278: I'm not sure if I'm reading it right but [Diff](https://github.com/filecoin-project/go-amt-ipld/blob/723fff1300c57c2f866505a7f3bd0796817e8bbf/diff.go#L42) doesn't seem to quickly check that the cids are equal and return immediately.  If this is right it is important for us to check equality quickly and set outArray to prevOutArray.  It might speed us up a lot on second run.

builtin/v9/migration/miner.go
```diff
@@ -165,3 +179,262 @@ func (m minerMigrator) migrateState(ctx context.Context, store cbor.IpldStore, i
...
+			}
+
+			outArray = prevOutSectors.Array
+		} else {
+			// first time we're doing this, do all the work
+			outArray, err = migrateSectorsFromScratch(ctx, store, inArray)
+			if err != nil {
+				return cid.Undef, xerrors.Errorf("failed to migrate sectors from scratch: %w", err)
+			}
+
```
ZenGround0, line 325: super nit: delete new line

builtin/v9/migration/miner.go
```diff
@@ -165,3 +179,262 @@ func (m minerMigrator) migrateState(ctx context.Context, store cbor.IpldStore, i
...
+	for i, c := range inDeadlines.Due {
+		if c == m.emptyDeadlineV8 {
+			outDeadlines.Due[i] = m.emptyDeadlineV9
+		} else {
+			var inDeadline miner8.Deadline
+			if err = store.Get(ctx, c, &inDeadline); err != nil {
+				return cid.Undef, err
+			}
+
+			outSectorsSnapshotCid, err := cache.Load(SectorsAmtKey(inDeadline.SectorsSnapshot), func() (cid.Cid, error) {
```
ZenGround0, line 378: If we are doing diffing in the SectorInfo context it probably makes sense to do diffing here too.  The function inside load should then be the extracted function used above.

builtin/v9/migration/miner.go
ZenGround0: Aha I'm missingt that this is inside cache.Load

manifest/manifest.go
```diff
@@ -12,6 +12,39 @@ import (
...
+		MinerKey,
+		MultisigKey,
+		PaychKey,
+		PowerKey,
+		RewardKey,
+		SystemKey,
+		VerifregKey,
+	}
+	return keys
+}
```
geoff-vball, line 46: We have a version on this in `chain/actors/manifest.go`. I think this is a better location for this, but it still should take actors version as an input.

builtin/v8/util/adt/set.go
arajasek: Yessir!

builtin/v9/migration/top.go
arajasek: Yessir!

builtin/v9/migration/miner.go
ZenGround0: And to be more clear I'm suggesting diffing against `MinerPrevSectorsInKey(minerAddr)`. We're never going to be more than 1 Day plus Wait time between premigrations of diffs.

If things are still taking a long time and you are feeling experimental you could introduce deadline index / address keys to do diffing against previously cached deadline sectors snapshots but I guess that this won't be a good complexity performance tradeoff.

builtin/v9/migration/miner.go
arajasek: Agreed. I'm gonna start an issue capturing possible further caching / speedups. I do want to keep the cache focused on highest impact stuff, because it can get big...

builtin/v9/migration/miner.go
arajasek: yeeee

builtin/v9/migration/miner.go
arajasek: I was torn about this, but I think I agree. Or at least don't feel strongly about it.

builtin/v9/migration/miner.go
arajasek: It's a little bit more complicated than that. The diffing for the Sectors AMT is based on snapshotting (for each miner) the input and output tree at the time of the premigration.

It's unclear what the equivalent to do here would be. We _could_ snapshot each deadline for each miner, but that feels like a lot. We could also just use the input/output of the migration of the "real" Sectors AMT (which we perform right before we start migrating deadlines) as the basis of the diff for the sectors snapshots in each of the deadlines. Realistically, that's as good a baseline for the snapshotted AMTs.

My thought was that all but 3-4 of the deadlines per miner will be perfectly cached by the premigration anyway (since their SectorsSnapshots can't change in the 2 hours between the premigration and the migration). Further, if they do change, there's a good chance the snapshot winds up being the same as the "real" Sectors AMT, and so are found in the cache from the previous step.

All that to say, I agree that some caching is possible here, but there's some open questions about the best way to do it. Some part of me is wary of introducing more caching than necessary lest doing so introduce bugs.

ZenGround0: > We could also just use the input/output of the migration of the "real" Sectors AMT (which we perform right before we start migrating deadlines) as the basis of the diff for the sectors snapshots in each of the deadlines. Realistically, that's as good a baseline for the snapshotted AMTs.

Yeah I think this is what I was trying to suggest with " diffing against MinerPrevSectorsInKey(minerAddr)" thought I'm missing some details.  Good point that "all but 3-4 of the deadlines per miner will be perfectly cached by the premigration anyway" so doing more caching is definitely not worth it.

builtin/v9/migration/verifreg.go
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		providerIDu64, err := address.IDFromAddress(providerIDAddress)
+		if err != nil {
+			return cid.Undef, nil, err
+		}
+
+		clientAllocationMap, ok := allocationsMapMap[clientIDAddress]
+		if !ok {
+			clientAllocationMap, err = adt9.AsMap(adtStore, emptyMapCid, builtin.DefaultHamtBitwidth)
```
ZenGround0, line 88: I don't think this adt.Map ever gets added to the allocationsMapMap golang map (https://go.dev/play/p/ouAvN7Ea813) causing all but the last iterated verified deal from client to be added.

ZenGround0 (commented):
A couple things before looking at top level

builtin/v9/migration/miner.go
```diff
@@ -165,3 +179,265 @@ func (m minerMigrator) migrateState(ctx context.Context, store cbor.IpldStore, i
...
+}
+
+func migrateSectorsWithCache(ctx context.Context, store adt8.Store, cache MigrationCache, minerAddr address.Address, inRoot cid.Cid) (cid.Cid, error) {
+	return cache.Load(SectorsAmtKey(inRoot), func() (cid.Cid, error) {
+		inArray, err := adt8.AsArray(store, inRoot, miner8.SectorsAmtBitwidth)
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("failed to read sectors array: %w", err)
+		}
+
+		okIn, prevInRoot, err := cache.Read(MinerPrevSectorsInKey(minerAddr))
```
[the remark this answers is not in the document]
arajasek: That...makes complete sense, yes.

builtin/v9/migration/miner.go
arajasek: Yeah, I'd prefer to leave that to a followup if that works for you -- see https://github.com/filecoin-project/go-state-types/issues/69.

builtin/v9/migration/top.go
```diff
@@ -135,30 +156,34 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
 	}
 
-	miner9Cid, ok := newManifest.Get("storageminer")
-	if !ok {
-		return cid.Undef, xerrors.Errorf("code cid for miner actor not found in new manifest")
-	}
-
 	migrations[systemActor.Code] = systemActorMigrator{newSystemCodeCID, newManifest.Data}
 
+	// The Miner Actor
```
ZenGround0, line 183: nit: move comment to code it refers to

builtin/v9/migration/top.go
```diff
@@ -307,7 +342,114 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
 
 	elapsed := time.Since(startTime)
 	rate := float64(doneCount) / elapsed.Seconds()
-	log.Log(rt.INFO, "All %d done after %v (%.0f/s). Flushing state tree root.", doneCount, elapsed, rate)
+	log.Log(rt.INFO, "All %d done after %v (%.0f/s). Starting deferred migrations.", doneCount, elapsed, rate)
+
+	// Fetch actor states needed for deferred migrations
```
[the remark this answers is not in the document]
ZenGround0: For threadsafety without hacking into the main wait group logic you can read inputs from actorsIn before spawning goroutine and write to actorsOut after all other migrations have finished.

builtin/v9/verifreg/verifreg_types.go
```diff
@@ -83,11 +84,21 @@ type FailCode struct {
 	Code ExitCode
 }
 
+// TODO: this shouldn't be here?
```
ZenGround0, line 87: Yeah what happens when we delete this?

ZenGround0 (approved):
I'm ok with this merging once correctness is addressed and defer performance to later.  However I would prefer getting at least the deferred migrations in parallel with the rest before merge since this looks like a great opportunity and in need of testing asap.

builtin/v9/migration/miner.go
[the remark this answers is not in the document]
arajasek: Ah, the limitation is that the cache (as currently written) can only have `value` of type `cid` (not `{cid, cid}`.

builtin/v9/migration/verifreg.go
```diff
@@ -0,0 +1,142 @@
...
+			}
+		}
+
+		if err = clientAllocationMap.Put(nextAllocationID, &verifreg9.Allocation{
+			Client:   abi.ActorID(clientIDu64),
+			Provider: abi.ActorID(providerIDu64),
+			Data:     proposal.PieceCID,
+			Size:     proposal.PieceSize,
+			TermMin:  proposal.Duration(),
+			TermMax:  market9.DealMaxDuration,
```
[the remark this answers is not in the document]
arajasek: Hmm, I implemented this based on the FIP https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0045.md#migration. Not sure what to do.

builtin/v9/migration/verifreg.go
arajasek: GOOD catch, thank you.

builtin/v9/migration/top.go
arajasek: This was intentionally placed, because the market state loading is for The Miner Actor. I can move it if it's only serving to confuse, though.

builtin/v9/migration/verifreg.go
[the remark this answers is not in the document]
ZenGround0: Offline I said its ok to keep as is.  One small change: I think you should set TermMax to DealMaxDuration + MarketDefaultAllocationTermBuffer.  This way deals with maximal deal lifetime get the same TermMax as they would from market code.

builtin/v9/migration/top.go
ZenGround0: ah ok in that case writing a bit more would also do it

builtin/v9/migration/miner.go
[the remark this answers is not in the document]
ZenGround0: Ah in that case this now makes complete sense

builtin/v9/migration/top.go
arajasek: Done

builtin/v9/migration/verifreg.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+
+	proposals, err := market8.AsDealProposalArray(adtStore, marketStateV8.Proposals)
+	if err != nil {
+		return cid.Undef, nil, xerrors.Errorf("failed to get proposals: %w", err)
+	}
+
+	pendingMap := make(map[abi.DealID]market8.DealProposal)
+	var proposal market8.DealProposal
+	if err = proposals.ForEach(&proposal, func(dealID int64) error {
```
anorth, line 34: This is a large map (every deal currently active). It would be much faster to iterate PendingProposals and look them up in Proposals.

builtin/v9/migration/verifreg.go
```diff
@@ -0,0 +1,143 @@
...
+	if err != nil {
+		return cid.Undef, nil, xerrors.Errorf("failed to flush allocations map: %w", err)
+	}
+
+	verifregStateV9 := verifreg9.State{
+		RootKey:                  verifregStateV8.RootKey,
+		Verifiers:                verifregStateV8.Verifiers,
+		RemoveDataCapProposalIDs: verifregStateV8.RemoveDataCapProposalIDs,
+		Allocations:              allocationsMapRoot,
+		NextAllocationId:         1,
```
anorth, line 133: ‼️ this should be `nextAllocationId`.

builtin/v9/migration/verifreg.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		if !ok {
+			return cid.Undef, nil, xerrors.New("failed to find client in init actor map")
+		}
+
+		clientIDu64, err := address.IDFromAddress(clientIDAddress)
+		if err != nil {
+			return cid.Undef, nil, err
+		}
```
anorth, line 70: Nit: there'd be good clarity gains from extracting pedestrian logic like this into functions so that the main migration code gets straight to the point.

builtin/v9/migration/verifreg.go
```diff
@@ -0,0 +1,143 @@
...
+			allocationsMapMap[clientIDAddress] = clientAllocationMap
+		}
+
+		if err = clientAllocationMap.Put(nextAllocationID, &verifreg9.Allocation{
+			Client:     abi.ActorID(clientIDu64),
+			Provider:   abi.ActorID(providerIDu64),
+			Data:       proposal.PieceCID,
+			Size:       proposal.PieceSize,
+			TermMin:    proposal.Duration(),
+			TermMax:    market9.DealMaxDuration,
```
anorth, line 102: ‼️ The max term should be a policy value _more than_ the deal duration, not a fixed value for all deals.

builtin/v9/verifreg/verified_registry_state.go
```diff
@@ -63,3 +75,151 @@ func ConstructState(store adt.Store, rootKeyAddress addr.Address) (*State, error
...
+	}
+
+	clientId, err := address.IDFromAddress(clientIdAddr)
+	if err != nil {
+		return nil, false, xerrors.Errorf("couldn't get ID from clientIdAddr: %s", clientIdAddr)
+	}
+
+	if uint64(allocation.Client) != clientId {
+		return nil, false, xerrors.Errorf("clientId: %d did not match client in allocation: %d", clientId, allocation.Client)
+	}
```
anorth, line 108: If the state somehow ended up like this, I don't think this is the right place to catch it. This does not mirror the Rust code.

builtin/v9/verifreg/verified_registry_state.go
```diff
@@ -63,3 +75,151 @@ func ConstructState(store adt.Store, rootKeyAddress addr.Address) (*State, error
...
+
+	providerId, err := address.IDFromAddress(providerIdAddr)
+	if err != nil {
+		return nil, false, xerrors.Errorf("couldn't get ID from providerIdAddr: %s", providerIdAddr)
+	}
+
+	if uint64(claim.Provider) != providerId {
+		return nil, false, xerrors.Errorf("providerId: %d did not match provider in claim: %d", providerId, claim.Provider)
+	}
```
anorth, line 142: Ditto. Less smarts in the state layer.

builtin/v9/verifreg/verified_registry_state.go
```diff
@@ -63,3 +75,151 @@ func ConstructState(store adt.Store, rootKeyAddress addr.Address) (*State, error
...
+	if found, err := actorToHamtMap.Get(key, &innerHamtCid); err != nil {
+		return cid.Undef, xerrors.Errorf("looking up key: %s: %w", key, err)
+	} else if !found {
+		return cid.Undef, xerrors.Errorf("did not find key: %s", key)
+	}
+
+	return cid.Cid(innerHamtCid), nil
+}
+
+func (st *State) AllocationsMap(store adt.Store, clientIdAddr address.Address) (map[AllocationId]Allocation, error) {
```
anorth, line 163: Nit: suggest AllocationsAsMap, or LoadAllocationsToMap or something else that suggests the amount of work this will do

anorth (changes requested):
I reviewed primarily for correctness of the state migrations themselves. I'm pretty far from the mechanics of how they are executed now and did not dive deep there.

builtin/v9/migration/verifreg.go
arajasek: @anorth I based this on the [text of the FIP ](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0045.md#migration). Am I misunderstanding something here?

anorth: You're right, I take it back.

arajasek: Zen had a slightly different suggestion [here](https://github.com/filecoin-project/go-state-types/pull/85#discussion_r988251213) -- do you have an opinion?

builtin/shared.go
```diff
@@ -10,6 +10,8 @@ import (
 // This value has been empirically chosen, but the optimal value for maps with different mutation profiles may differ.
 const DefaultHamtBitwidth = 5
 
+const DefaultTokenActorBitwidth = 3
```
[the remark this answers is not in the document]
arajasek: @anorth It's also needed at other times when the datacap actor state is constructed, eg. when creating new genesis blocks, starting directly from v9 actors.

builtin/v9/verifreg/verified_registry_state.go
```diff
@@ -14,6 +16,16 @@ import (
...
 
+var DataCapGranularity = builtin.TokenPrecision
+
+func DataCapToTokens(d DataCap) abi.TokenAmount {
+	return big.Mul(d, DataCapGranularity)
+}
+
+func TokensToDatacap(t abi.TokenAmount) DataCap {
+	return big.Div(t, DataCapGranularity)
+}
```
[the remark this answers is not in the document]
arajasek: Good call.

builtin/v9/migration/verifreg.go
anorth: Wait, yeah I un-take it back. We want the same behaviour as the market would have done. If a client set 540d deal term, the max should be +buffer on that.

arajasek: Good, thanks! I will implement this change and update the FIP for overall Zen.

builtin/v9/migration/verifreg.go
anorth: Oh right, I see. There's good reason for the set of CIDs.

builtin/v9/migration/verifreg.go
arajasek: Holy fork, thank you!

builtin/v9/verifreg/verified_registry_state.go
arajasek: I'd prefer to leave it -- I think it's reasonable for the state layer to know that if it's asked to find a claim for a specific provider, then the returned claim should match that provider.

No harm done in the sanity check, I think.

builtin/v9/migration/verifreg.go
```diff
@@ -99,14 +76,19 @@ func migrateVerifreg(ctx context.Context, adtStore adt8.Store, priorEpoch abi.Ch
...
+
 		if err = clientAllocationMap.Put(nextAllocationID, &verifreg9.Allocation{
 			Client:     abi.ActorID(clientIDu64),
 			Provider:   abi.ActorID(providerIDu64),
 			Data:       proposal.PieceCID,
 			Size:       proposal.PieceSize,
 			TermMin:    proposal.Duration(),
-			TermMax:    market9.DealMaxDuration,
-			Expiration: verifreg9.MaximumVerifiedAllocationExpiration + priorEpoch,
+			TermMax:    market9.DealMaxDuration + market9.MarketDefaultAllocationTermBuffer,
```
anorth, line 90: I think we want `proposal.Duration() + market9.MarketDefaultAllocationTermBuffer`

anorth (approved):

ZenGround0 (approved):

PR #88, opened 2022-10-05
Add accessors in verifreg actor

geoff-vball:
Adds accessor functions that return a map of all claims or all allocations for a given actor.

### DIFF
--- a/builtin/v9/verifreg/verified_registry_state.go
+++ b/builtin/v9/verifreg/verified_registry_state.go
@@ -144,18 +144,82 @@ func (st *State) FindClaim(store adt.Store, providerIdAddr address.Address, clai
 	return &claim, true, nil
 }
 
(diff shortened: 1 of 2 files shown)


builtin/v9/verifreg/verified_registry_state.go
```diff
@@ -144,18 +144,82 @@ func (st *State) FindClaim(store adt.Store, providerIdAddr address.Address, clai
...
+	if err != nil {
+		return nil, xerrors.Errorf("couldn't get map: %x", err)
+	}
+
+	var goMap = make(map[ClaimId]Claim)
+	var out Claim
+	err = adtMap.ForEach(&out, func(key string) error {
+		uintKey, err := abi.ParseUIntKey(key)
+		if err != nil {
+			return xerrors.Errorf("couldn't parse key to uint: %x", err)
```
arajasek, line 215:
```suggestion
			return xerrors.Errorf("couldn't parse key to uint: %w", err)
```

arajasek (approved):

PR #89, opened 2022-10-07
Make migration deterministic

arajasek:
Three fixes:

- Make the migration deterministic, by eliminating iterating over go maps
- Cache the datacap actor result based on the _verified registry's_ head. If this head changes we need to redo the migration, if not the cached result is good.
- Create the datacap actor in the _input_ actors tree, not output, so that it actually goes through a migration (this was a bug introduced through copypasta).

### DIFF
--- a/builtin/v9/migration/market.go
+++ b/builtin/v9/migration/market.go
@@ -9,21 +9,20 @@ import (
 	adt8 "github.com/filecoin-project/go-state-types/builtin/v8/util/adt"
 	market9 "github.com/filecoin-project/go-state-types/builtin/v9/market"
 	adt9 "github.com/filecoin-project/go-state-types/builtin/v9/util/adt"
-	verifreg9 "github.com/filecoin-project/go-state-types/builtin/v9/verifreg"
 	"github.com/ipfs/go-cid"
 	typegen "github.com/whyrusleeping/cbor-gen"
 	"golang.org/x/xerrors"
 )
 
-func migrateMarket(ctx context.Context, adtStore adt8.Store, dealsToAllocations map[abi.DealID]verifreg9.AllocationId, marketStateV8 market8.State, emptyMapCid cid.Cid) (cid.Cid, error) {
+func migrateMarket(ctx context.Context, adtStore adt8.Store, dealAllocationTuples []DealAllocationTuple, marketStateV8 market8.State, emptyMapCid cid.Cid) (cid.Cid, error) {
 	pendingDealAllocationIdsMap, err := adt9.AsMap(adtStore, emptyMapCid, builtin.DefaultHamtBitwidth)
 	if err != nil {
 		return cid.Undef, xerrors.Errorf("failed to load empty map: %w", err)
 	}
 
-	for dealID, allocationID := range dealsToAllocations {
-		cborAllocationID := typegen.CborInt(allocationID)
-		if err = pendingDealAllocationIdsMap.Put(abi.UIntKey(uint64(dealID)), &cborAllocationID); err != nil {
+	for _, tuple := range dealAllocationTuples {
+		cborAllocationID := typegen.CborInt(tuple.Allocation)
+		if err = pendingDealAllocationIdsMap.Put(abi.UIntKey(uint64(tuple.Deal)), &cborAllocationID); err != nil {
 			return cid.Undef, xerrors.Errorf("failed to populate pending deal allocations map: %w", err)
 		}
 	}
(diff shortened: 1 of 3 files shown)


builtin/v9/migration/market.go
```diff
@@ -9,21 +9,20 @@ import (
...
 	if err != nil {
 		return cid.Undef, xerrors.Errorf("failed to load empty map: %w", err)
 	}
 
-	for dealID, allocationID := range dealsToAllocations {
-		cborAllocationID := typegen.CborInt(allocationID)
-		if err = pendingDealAllocationIdsMap.Put(abi.UIntKey(uint64(dealID)), &cborAllocationID); err != nil {
+	for _, tuple := range dealAllocationTuples {
+		cborAllocationID := typegen.CborInt(tuple.Allocation)
+		if err = pendingDealAllocationIdsMap.Put(abi.UIntKey(uint64(tuple.Deal)), &cborAllocationID); err != nil {
```
ZenGround0, line 25: Did you observe non determinism from inserting into the HAMT in different orders?

builtin/v9/migration/top.go
```diff
@@ -244,10 +244,12 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
 
-	if err = actorsOut.SetActor(builtin.DatacapActorAddr, &Actor{
+	if err = actorsIn.SetActor(builtin.DatacapActorAddr, &Actor{
 		Code: dataCapCode,
-		// we just need to put _something_ defined, this never gets read
-		Head:       emptyMapCid,
+		// we supply the verified registry's head here for caching purposes
+		// if the verifreg head has changed, a fresh migration is needed
+		// if not, the cached datacap migration can be used
+		Head:       verifregActorV8.Head,
```
ZenGround0, line 252: I don't understand how the regular [cached migration](https://github.com/filecoin-project/go-state-types/pull/89/files#diff-0f71e4e5376421454dd6e1792714840e621766fb31e279de1749157bf59a8bd3R554) is going to map this to the correct thing in the case of a cache hit.  Why isn't it mapping cached verifreg in to verifreg out?  Where is the logic to swap the verifreg state for datacap state?

ZenGround0: Despite whatever I'm missing that makes the above work this approach is too clever for its own good.  cachedMigration does one simple thing which is distinct from what we want here.  I think you want to define a new migration transform like cachedMigration that explicitly checks if verifreg is in the cache and then migrates datacap.

ZenGround0 (changes requested):

builtin/v9/migration/market.go
arajasek: Yes, I did. That is...bad.

builtin/v9/migration/top.go
arajasek: TL;DR: I agree with you, we can just drop the caching here -- it's a little sad because we will very likely have a cache hit (the verifreg changes infrequently), but it's likely not the critical path and not worth a bug.

The caching, as I understand it, is keyed by an `Address, Head` tuple. During the premigration, we map `Address, Head` to the `MigratedHead`. If at the time of the actual migration we have a hit for `Address, Head`, we blindly use `MigratedHead`.

The DataCap actor is in the peculiar position of being loadable from the cache if the verifreg head doesn't change. Thus, if we enter `DataCapAddr, VerifregHead` into our cache, mapping it to `NewDataCapHead`, we'll correctly load it. 

Note that we will also use the cached entry for the Verifreg migration, which will have been entered with a key of `VerifregAddr, VerifregHead`. 

But, all in all, this is janky. Let's see if we can make do with no caching for the DataCap actor?

arajasek:
Pushed a commit killing DataCap caching.

builtin/v9/migration/top.go
ZenGround0: I think you should make use of this caching I just want it done in a way that removes faking out the datacap actor's head cid.  If you implement a variation on `cachedMigration` then this would be nice.  I.e. define `conditionalMigrator` that tracks two addresses and checks the cache for the first one before checking the cache for the second one.

ZenGround0 (approved):
Approved with suggestion on how to regain the property from the first one with less tricky code

builtin/v9/migration/market.go
Kubuxu: That shouldn't be a thing.

PR #95, opened 2022-10-12
feat: Datacap methods return Cbor-able strings

geoff-vball:
Implements CborString and uses them in Datacap method returns

### DIFF
--- a/abi/cbor_string.go
+++ b/abi/cbor_string.go
@@ -1,0 +1,46 @@
+package abi
+
+import (
+	"io"
+
+	"golang.org/x/xerrors"
+
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+type CborString string
+
+func (t *CborString) MarshalCBOR(w io.Writer) error {
+	scratch := make([]byte, 8)
+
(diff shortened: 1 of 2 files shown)


abi/cbor_string.go
```diff
@@ -0,0 +1,46 @@
...
+
+import (
+	"io"
+
+	"golang.org/x/xerrors"
+
+	cbg "github.com/whyrusleeping/cbor-gen"
+)
+
+type CborString string
```
ZenGround0, line 11: If its possible it would be better to gen this directly (updates to cbor gen propagate, no need to review independently for correctness).  Does that not work with cbg?  If so it seems like its worth filing an issue over there.

geoff-vball: Cbor-gen only works with structs, not primitives or type definitions thereof. There are hardcoded utilities for some primitives like `[CborInt](https://github.com/whyrusleeping/cbor-gen/blob/f29c37e9c44c4a318d70f78686893ec960cfa771/utils.go#L712)` in the library, but there doesn't seem to be one for `string`

ZenGround0 (approved):

PR #97, opened 2022-10-17
fix: verifreg migration: set TermMax to DealMaxDuration

arajasek:
This un-un-takes back the discussion [here](https://github.com/filecoin-project/go-state-types/pull/85#discussion_r989515121), and undoes the change in #91. 

If we land this, we should close https://github.com/filecoin-project/FIPs/pull/485, since we would be agreeing that the FIP is correct after all.


ZenGround0 (approved):
LGTM

PR #98, opened 2022-10-17
Export actor method names

frrist:


frrist:
lint failures are due to unkeyed fields in the MethodMeta struct. I think ignoring these linter errors would be preferred, as leaving the declaration of the structure unkey will ensure the package fails to build if new fields are added to the MethodMeta struct - serving as a reminder to the implementor when making changes.

arajasek:
@frrist Thanks for the PR!

I agree, let's ignore those lint errors.

@geoff-vball also pointed out that this is a good opportunity to drop the `abi.EmptyValue` param from every method -- this is a placeholder for the Runtime interface that all methods receive as their first param. From the perspective of a user, however, this is irrelevant -- we could simply reflect on the 0th param in places like [this](https://github.com/filecoin-project/lotus/blob/94add978b01f365fba1c838a61497ac035f8095b/chain/vm/invoker.go#L143).

arajasek:
But that can be a future thing -- I'm happy to land this as-is after the linter ignore is added.

frrist:
Sounds good; I'll leave off the EmptyValue change. I have addressed the linter errors with a config file change as well.

arajasek (approved):

PR #101, opened 2022-10-20
add the Ethereum Address Manager as a singleton.

raulk:
Need the actor ID separately because we don't want to parse it from the address every time we construct a delegated address.


raulk:
Self-merging to unblock myself. Can always adjust on the branch or before it lands on master.

PR #102, opened 2022-10-20
add the exchange types for the EAM.

raulk:
I know the CreateAccount method is not implemented and the exchange type may change, but easier to just add it now and adjust later, for parity with what's sitting in bulitin-actors/next.


raulk:
Self-merging to unblock myself. Can always adjust on the branch or before it lands on master.

PR #108, opened 2022-10-21
Market migration: Correctly migrate deal states

arajasek:
Quoting [the FIP ](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0045.md#migration),

> Add the pending allocation mapping to market state, and claim ID field to deal states (all 0)

The second half of this direction (set claim ID field of deal states to 0) was missing. This PR adds the necessary operation.


magik6k (approved):
One slight concern about perf, do we know how long it takes on mainnet with its 13M deals?

ZenGround0 (approved):

ZenGround0:
> One slight concern about perf, do we know how long it takes on mainnet with its 13M deals?

AMT diffing used for sectors AMT should also help here

PR #110, opened 2022-10-21
Add verifreg balance to datacap

ZenGround0:
This refactors full proposal traversal to a single preliminary call that collects deal IDs of verified pending deals for verifreg migration and their total size for datacap migration

### DIFF
--- a/builtin/v9/migration/verifreg.go
+++ b/builtin/v9/migration/verifreg.go
@@ -25,11 +25,7 @@ type DealAllocationTuple struct {
 	Allocation verifreg9.AllocationId
 }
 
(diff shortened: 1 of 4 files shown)


builtin/v9/migration/verifreg.go
```diff
@@ -39,38 +35,23 @@ func migrateVerifreg(ctx context.Context, adtStore adt8.Store, priorEpoch abi.Ch
...
-		if err != nil {
-			return xerrors.Errorf("failed to check pending: %w", err)
-		}
 
-		// Nothing to do for not-pending deals
-		if !isPending {
-			return nil
+	for _, dealID := range pendingDeals {
+		proposal, err := proposals.GetDealProposal(dealID)
+		if err != nil || proposal == nil {
```
ZenGround0, line 41: proposal == nil should not be reachable but I figured an explicit error is minutely nicer than a panic

arajasek (approved):

builtin/v9/migration/verifreg.go
arajasek: Agreed

PR #118, opened 2022-11-22
chore: Move MinimumVerifiedAllocationSize to network.go

geoff-vball:
Mainnet params all live in network.go. MinimumVerifiedAllocationSize is a configurable network param, so it should live there too.

Also rename it to MinVerifiedDealSize to match with lotus params. We now use the terms allocation and claim, but I think VerfiedDeal is still a good catch-all term for these.


arajasek (approved):

arajasek:
I'm okay with this, but we could also just make this a private, internal field that only the invariants checker uses.

PR #119, opened 2022-11-22
fix: Universal Receiver method for account and verifreg take abi.EmptyValue

geoff-vball:
Lotus needs a cbor-marshallable data type to pass in here. Both the `Account` and `Multisig` actors ignore the params passed in to `UniversalReceiverHook`, so we will pass `*abi.EmptyValue`.

### DIFF
--- a/builtin/v9/account/methods.go
+++ b/builtin/v9/account/methods.go
@@ -11,5 +11,5 @@ var Methods = map[uint64]builtin.MethodMeta{
 	1: {"Constructor", *new(func(*address.Address) *abi.EmptyValue)},                   // Constructor
 	2: {"PubkeyAddress", *new(func(*abi.EmptyValue) *address.Address)},                 // PubkeyAddress
 	3: {"AuthenticateMessage", *new(func(*AuthenticateMessageParams) *abi.EmptyValue)}, // AuthenticateMessage
-	uint64(builtin.UniversalReceiverHookMethodNum): {"UniversalReceiverHook", *new(func(*[]byte) *abi.EmptyValue)}, // UniversalReceiverHook
+	uint64(builtin.UniversalReceiverHookMethodNum): {"UniversalReceiverHook", *new(func(*abi.CborBytesTransparent) *abi.EmptyValue)}, // UniversalReceiverHook
 }
--- a/abi/cbor_bytes_transparent.go
+++ b/abi/cbor_bytes_transparent.go
@@ -1,0 +1,20 @@
+package abi
+
+import (
+	"io"
+)
+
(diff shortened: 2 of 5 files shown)


anorth:
I don't quite understand the context here, but not that EmptyValue would fail to deserialise if a non-empty value was passed, and we'll definitely pass non-empty values here on chain.

frrist:
One problem is Lotus expects all message params and returns to be CBOR-able, when they are not [methods such as these](https://github.com/filecoin-project/lotus/blob/13b59c9c23d31f45babe573ac4f0a366e2c881de/chain/stmgr/utils.go#L30-L50) can panic. One alternative to this change could be to modify the methods in Lotus instead.

builtin/v9/account/methods.go
```diff
@@ -11,5 +11,5 @@ var Methods = map[uint64]builtin.MethodMeta{
 	1: {"Constructor", *new(func(*address.Address) *abi.EmptyValue)},                   // Constructor
 	2: {"PubkeyAddress", *new(func(*abi.EmptyValue) *address.Address)},                 // PubkeyAddress
 	3: {"AuthenticateMessage", *new(func(*AuthenticateMessageParams) *abi.EmptyValue)}, // AuthenticateMessage
-	uint64(builtin.UniversalReceiverHookMethodNum): {"UniversalReceiverHook", *new(func(*[]byte) *abi.EmptyValue)}, // UniversalReceiverHook
+	uint64(builtin.UniversalReceiverHookMethodNum): {"UniversalReceiverHook", *new(func(*abi.EmptyValue) *abi.EmptyValue)}, // UniversalReceiverHook
```
frrist, line 14: I think `CborBytes` would be _better_ but still not quite right since (I think) it's a not a requirement that these params are CBOR-able.

geoff-vball:
Updated this with an implementation of `CborBytes`, but I'm not entirely convinced it's the correct solution yet.

arajasek:
> Updated this with an implementation of `CborBytes`, but I'm not entirely convinced it's the correct solution yet.

@geoff-vball You're right, CBOR bytes definitely isn't the right thing to use here. These really can be any bytes, so I think we want CBOR transparent as we hacked up on Slack:

```
type CborBytesTransparent []byte

func (t *CborBytesTransparent) MarshalCBOR(w io.Writer) error {
	_, err := w.Write(*t)
	return err
}

func (t *CborBytesTransparent) UnmarshalCBOR(r io.Reader) error {
	_, err := r.Read(*t)
	return err
}
```

It's fine if it's imperfect.

`abi.EmptyValue` is overly hacky.

geoff-vball:
@arajasek Updated with the transparent bytes. I liked your original name of `CborBytesTransparent` so that's what I'm going with.

arajasek (approved):

PR #121, opened 2022-11-23
Update version.go

jennijuju:


arajasek (approved):

arajasek:
I'm not sure how much I agree with this, given that the v10 state types (corresponds to network v18) is specifically the code needed for FVM 2.1. But I do think on-balance it's better to wait to label it as such in this file until things are official.

(And def agreed that upgrade name should wait until it's confirmed by FF)

Issue #127, opened 2022-12-08
Migrate actors to v5


jennijuju:

geoff-vball:
ActorV5 is already implemented in `actor_tree.go`. We just need to calculate and store the predictable address for any actors that we want to migrate. 

@Stebalien can probably define what actor types need to be migrated here.

Stebalien:
The answer is "none". We've decided to not store _any_ predictable addresses, only f4 addresses. I'm currently updating FIP0048 to include this (instead of writing an entirely new FIP).

Stebalien:
See https://github.com/filecoin-project/ref-fvm/pull/1290.

PR #136, opened 2022-12-15
feat: Implement barebones v10 migration

geoff-vball:
Extracts some common migration functionality out of v10. Migrates code CIDs and the system actor. Putting this here as a starting point as I already implemented it for FIPI-0047

Also adds nil address in ActorV5 which Closes https://github.com/filecoin-project/go-state-types/issues/127

### DIFF
--- a/builtin/v10/migration/top.go
+++ b/builtin/v10/migration/top.go
@@ -1,0 +1,269 @@
+package migration
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/builtin"
+	system9 "github.com/filecoin-project/go-state-types/builtin/v9/system"
+	adt9 "github.com/filecoin-project/go-state-types/builtin/v9/util/adt"
+	"github.com/filecoin-project/go-state-types/manifest"
+	"github.com/filecoin-project/go-state-types/migration"
+	"github.com/filecoin-project/go-state-types/rt"
+
+	"github.com/ipfs/go-cid"
+	cbor "github.com/ipfs/go-ipld-cbor"
+	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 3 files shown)


builtin/v10/migration/top.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// Maps prior version code CIDs to migration functions.
+	migrations := make(map[cid.Cid]migration.ActorMigration)
+	// Set of prior version code CIDs for actors to defer during iteration, for explicit migration afterwards.
+	deferredCodeIDs := make(map[cid.Cid]struct{})
+
+	// Populated from oldManifestData
+	oldCodeIDMap := make(map[string]cid.Cid, len(oldManifestData.Entries))
+
+	miner9Cid := cid.Undef
```
arajasek, line 81: We shouldn't need to special-case the miner CID yet (and likely won't have to at all since FIP-0047 is likely moving to nv19).

builtin/v10/migration/top.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// Set up new actor record with the migrated state.
+	return &migrationJobResult{
+		job.Address, // Unchanged
+		builtin.ActorV5{
+			Code:       result.NewCodeCID,
+			Head:       result.NewHead,
+			CallSeqNum: job.ActorV4.CallSeqNum, // Unchanged
+			Balance:    job.ActorV4.Balance,    // Unchanged
+			Address:    nil,                    // TODO input predictable address for accounts
```
arajasek, line 277: The TODO can stay, but I _think_ the current proposal is to not actually populate this for accounts.

arajasek (approved):
Thanks for this! Just the one nit about the miner CID.

builtin/v10/migration/top.go
geoff-vball: Good catch!

builtin/v10/migration/top.go
geoff-vball: Yeah, I don't think we have anything to do here, but will keep TODO to check later.

PR #139, opened 2022-12-21
Init: Implement new MapAddressToNewID method

arajasek:
Matches behaviour introduced in https://github.com/filecoin-project/builtin-actors/pull/950


arajasek:
@Stebalien wants this method to die, so I'll do that instead...

PR #142, opened 2022-12-22
feat: implement the v10 migration

arajasek:
- Create the EAM Actor with "empty object" head
- Create the EthZeroAddress actor as an EthAccount, map it in the init actor (without a robust address)
- everything else is just a CodeID update, and has the new `Address` field nil

### DIFF
--- a/builtin/v10/migration/top.go
+++ b/builtin/v10/migration/top.go
@@ -6,6 +6,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	init10 "github.com/filecoin-project/go-state-types/builtin/v10/init"
+
 	"github.com/filecoin-project/go-address"
 	"github.com/filecoin-project/go-state-types/abi"
 	"github.com/filecoin-project/go-state-types/builtin"
@@ -75,6 +77,7 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID 
 	// Set of prior version code CIDs for actors to defer during iteration, for explicit migration afterwards.
 	deferredCodeIDs := make(map[cid.Cid]struct{})
 
+	oldInitCodeCID := cid.Undef
 	for _, oldEntry := range oldManifestData.Entries {
 		newCodeCID, ok := newManifest.Get(oldEntry.Name)
 		if !ok {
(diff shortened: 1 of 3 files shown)


builtin/v10/migration/top.go
```diff
@@ -221,6 +243,66 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
+	}
+
+	// Create the EAM Actor
+
+	eamAct, err := CreateEAMActor(&newManifest, emptyObj)
+	if err != nil {
+		return cid.Undef, xerrors.Errorf("failed to create the EAM Actor: %w", err)
+	}
+
+	eamAddr, err := address.NewIDAddress(builtin.EthereumAddressManagerActorID)
```
geoff-vball, line 260: I would use `builtin.EthereumAddressManagerActorAddr`

geoff-vball (approved):
Approved with one nit

builtin/v10/migration/top.go
arajasek: Yessir!

builtin/v10/migration/top.go
```diff
@@ -75,25 +77,45 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
+	if !ok {
+		return cid.Undef, xerrors.Errorf("code cid for init actor not found in new manifest")
+	}
+
+	ethZeroAddr, err := makeEthZeroAddress()
+	if err != nil {
+		return cid.Undef, xerrors.Errorf("failed to make eth zero address: %w", err)
+	}
+
+	migrations[oldInitCodeCID] = initActorMigrator{OutCodeCID: newInitCodeCID, EthZeroAddress: ethZeroAddr}
```
Stebalien, line 118: Oh, I see. Yeah.... that's not so fun. Could we just apply it as an implicit message? Or is that not allowed.

builtin/v10/migration/top.go
```diff
@@ -221,6 +243,61 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
+	eamAct, err := CreateEAMActor(&newManifest, emptyObj)
+	if err != nil {
+		return cid.Undef, xerrors.Errorf("failed to create the EAM Actor: %w", err)
+	}
+
+	if err := actorsOut.SetActorV5(builtin.EthereumAddressManagerActorAddr, eamAct); err != nil {
+		return cid.Undef, xerrors.Errorf("failed to set EAM Actor: %w", err)
+	}
+
+	// Create the EthZeroAddress as an EthAccount
```
Stebalien, line 264: I'm not seeing us _create_ it. Are we just checking to see if it exists?

arajasek: WDYM? There's basically 2 steps to "creating" it:

- Assigning it an ID in the init actor (happens as part of the init actor migration)
- Adding that ID to the state tree as an EthAccount, which happens ~25 lines after this comment.

We're _not_ assigning it a robust address, but that seemed okay to me?

builtin/v10/migration/top.go
Stebalien: Fair enough.

builtin/v10/migration/top.go
Stebalien: > Assigning it an ID in the init actor (happens as part of the init actor migration)

Ah, this is what I was looking for. You're right, we're creating the actor here, we've already assigned the address.

PR #143, opened 2023-01-02
Add new methods and types for FEVM

arajasek:
New types, methods, and actors that have been added to the built-in actors.


abi/cbor_bytes.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+	"io"
+
+	cbg "github.com/whyrusleeping/cbor-gen"
+	"golang.org/x/xerrors"
+)
+
+type CborBytes []byte
```
Stebalien, line 11: We should consider putting this into cbor-gen (and using cborgen's ReadByteArray/WriteByteArray.

abi/cbor_bytes.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (t *CborBytes) UnmarshalCBOR(r io.Reader) error {
+
+	br := cbg.GetPeeker(r)
+	maj, extra, err := cbg.CborReadHeader(br)
+	if err != nil {
+		return err
+	}
+
+	if extra > cbg.ByteArrayMaxLen {
```
Stebalien, line 34: This is probably too small (e.g., it won't fit an entire smart contract).

builtin/methods.go
```diff
@@ -386,3 +386,9 @@ var MethodsEAM = struct {
 	2,
 	3,
 }
+
```
geoff-vball, line 389: This file needs embryo methods

arajasek: It doesn't have any methods (not even a constructor), but I'll add an empty list here.

abi/cbor_bytes.go
arajasek: Hmm, if so we have a more serious problem -- this limit applies to pretty much every type in Filecoin (as implemented by Golang clients), including the `Message` itself. That means Lotus won't consider such messages valid.

(what happens here is less important since it's only used for UX)

abi/cbor_bytes.go
arajasek: ...future...improvement...

arajasek: https://github.com/whyrusleeping/cbor-gen/issues/75

geoff-vball (approved):

abi/cbor_bytes.go
Stebalien: Steven is confused. Steven will go back to the caverns of rust.

(the max here is ~2MiB and isn't a problem)

PR #148, opened 2023-01-12
Add evm state fields and invariants for v10

shrenujbansal:
- Add new EVM actor state fields: Bytecode_hash and Tombstone
- Add invariants for new actor types and update existing invariants required for v10 changes

### DIFF
--- a/builtin/v10/check.go
+++ b/builtin/v10/check.go
@@ -58,8 +58,11 @@ func CheckStateInvariants(tree *builtin.ActorTree, priorEpoch abi.ChainEpoch, ac
 		}
 		totalFIl = big.Add(totalFIl, actor.Balance)
 
-		if actor.Address != nil && actor.Address.Protocol() == address.Delegated {
-			delegatedAddrs = append(delegatedAddrs, *actor.Address)
+		if actor.Address != nil {
+			acc.Require(actor.Address.Protocol() == address.Delegated, "actor.Address %v is not a delegated address", *actor.Address)
+			if actor.Address.Protocol() == address.Delegated {
+				delegatedAddrs = append(delegatedAddrs, *actor.Address)
+			}
 		}
 
 		switch actor.Code {
--- a/builtin/v10/check.go
+++ b/builtin/v10/check.go
@@ -1,0 +1,359 @@
+package v10
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/filecoin-project/go-state-types/builtin/v10/datacap"
+	"github.com/filecoin-project/go-state-types/builtin/v10/evm"
+
+	"github.com/filecoin-project/go-state-types/manifest"
+	"github.com/ipfs/go-cid"
+
+	"github.com/filecoin-project/go-state-types/builtin/v10/account"
+	"github.com/filecoin-project/go-state-types/builtin/v10/cron"
+	init_ "github.com/filecoin-project/go-state-types/builtin/v10/init"
+	"github.com/filecoin-project/go-state-types/builtin/v10/market"
+	"github.com/filecoin-project/go-state-types/builtin/v10/miner"
+	"github.com/filecoin-project/go-state-types/builtin/v10/multisig"
+	"github.com/filecoin-project/go-state-types/builtin/v10/paych"
+	"github.com/filecoin-project/go-state-types/builtin/v10/power"
+	"github.com/filecoin-project/go-state-types/builtin/v10/reward"
+	"github.com/filecoin-project/go-state-types/builtin/v10/verifreg"
+
+	"github.com/filecoin-project/go-state-types/builtin"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-state-types/abi"
+	"github.com/filecoin-project/go-state-types/big"
+	"golang.org/x/xerrors"
+)
+
+// Within this code, Go errors are not expected, but are often converted to messages so that execution
+// can continue to find more errors rather than fail with no insight.
+// Only errors thar are particularly troublesome to recover from should propagate as Go errors.
+func CheckStateInvariants(tree *builtin.ActorTree, priorEpoch abi.ChainEpoch, actorCodes map[string]cid.Cid) (*builtin.MessageAccumulator, error) {
+	acc := &builtin.MessageAccumulator{}
+	totalFIl := big.Zero()
+	var initSummary *init_.StateSummary
+	var cronSummary *cron.StateSummary
+	var verifregSummary *verifreg.StateSummary
+	var datacapSummary *datacap.StateSummary
+	var marketSummary *market.StateSummary
+	var rewardSummary *reward.StateSummary
+	var accountSummaries []*account.StateSummary
+	var powerSummary *power.StateSummary
+	var paychSummaries []*paych.StateSummary
+	var multisigSummaries []*multisig.StateSummary
+	minerSummaries := make(map[address.Address]*miner.StateSummary)
+
+	if err := tree.ForEachV5(func(key address.Address, actor *builtin.ActorV5) error {
+		acc := acc.WithPrefix("%v ", key) // Intentional shadow
+		if key.Protocol() != address.ID {
+			acc.Addf("unexpected address protocol in state tree root: %v", key)
+		}
+		totalFIl = big.Add(totalFIl, actor.Balance)
+
+		switch actor.Code {
+		case actorCodes[manifest.SystemKey]:
+
+		case actorCodes[manifest.InitKey]:
+			var st init_.State
+			if err := tree.Store.Get(tree.Store.Context(), actor.Head, &st); err != nil {
+				fmt.Println("init invariant error = ", err)
+				return err
+			}
+			summary, msgs := init_.CheckStateInvariants(&st, tree, actorCodes)
+			acc.WithPrefix("init: ").AddAll(msgs)
+			initSummary = summary
+		case actorCodes[manifest.CronKey]:
+			var st cron.State
+			if err := tree.Store.Get(tree.Store.Context(), actor.Head, &st); err != nil {
+				return err
+			}
+			summary, msgs := cron.CheckStateInvariants(&st, tree.Store)
+			acc.WithPrefix("cron: ").AddAll(msgs)
+			cronSummary = summary
+		case actorCodes[manifest.AccountKey]:
+			var st account.State
+			if err := tree.Store.Get(tree.Store.Context(), actor.Head, &st); err != nil {
+				return err
+			}
+			summary, msgs := account.CheckStateInvariants(&st, key)
+			acc.WithPrefix("account: ").AddAll(msgs)
+			accountSummaries = append(accountSummaries, summary)
+		case actorCodes[manifest.PowerKey]:
+			var st power.State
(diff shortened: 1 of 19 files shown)


builtin/v10/check.go
```diff
@@ -0,0 +1,359 @@
...
+			if err := tree.Store.Get(tree.Store.Context(), actor.Head, &st); err != nil {
+				return err
+			}
+			summary, msgs := datacap.CheckStateInvariants(&st, tree.Store)
+			acc.WithPrefix("datacap: ").AddAll(msgs)
+			datacapSummary = summary
+		case actorCodes[manifest.EvmKey]:
+			var st evm.State
+			if err := tree.Store.Get(tree.Store.Context(), actor.Head, &st); err != nil {
+				fmt.Println("evm invariant error = ", err)
```
shrenujbansal, line 151: Remove this

builtin/v10/check.go
```diff
@@ -0,0 +1,359 @@
...
+		pendingAllocationsTotal = big.Add(pendingAllocationsTotal, big.NewIntUnsigned(uint64(allocation.Size)))
+	}
+
+	pendingAllocationsTotal = big.Mul(pendingAllocationsTotal, verifreg.DataCapGranularity)
+	verifregId, err := address.IDFromAddress(builtin.VerifiedRegistryActorAddr)
+	acc.RequireNoError(err, "could not get verifreg ID from address")
+	verifregBalance, found := datacapSummary.Balances[abi.ActorID(verifregId)]
+	if !found {
+		verifregBalance = big.Zero()
+	}
```
shrenujbansal, line 306: above 3 lines added in comparison to v9

builtin/v10/init/invariants.go
```diff
@@ -0,0 +1,80 @@
...
+
+		foundAddr, found := reverse[actorId]
+		isPair := (keyAddr.Protocol() == addr.Actor && foundAddr.Protocol() == addr.Delegated) ||
+			(keyAddr.Protocol() == addr.Delegated && foundAddr.Protocol() == addr.Actor)
+		dup := found && !isPair
+		acc.Require(!dup, "duplicate mapping to ID %v: %v, %v", actorId, keyAddr, foundAddr)
+		reverse[actorId] = keyAddr
+
+		initSummary.AddrIDs[keyAddr] = actorId
+
```
shrenujbansal, line 61: everything below newly added to init invariants

builtin/v10/init/invariants.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+
+		acc.Require(keyAddr.Protocol() != addr.ID, "key %v is an ID address", keyAddr)
+		acc.Require(keyAddr.Protocol() <= addr.Delegated, "unknown address protocol for key %v", keyAddr)
+		acc.Require(actorId >= builtin.FirstNonSingletonActorId, "unexpected singleton ID value %v", actorId)
+
+		foundAddr, found := reverse[actorId]
+		isPair := (keyAddr.Protocol() == addr.Actor && foundAddr.Protocol() == addr.Delegated) ||
+			(keyAddr.Protocol() == addr.Delegated && foundAddr.Protocol() == addr.Actor)
+		dup := found && !isPair
```
shrenujbansal, line 56: newly added

builtin/v10/init/invariants.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/filecoin-project/go-state-types/builtin/v10/util/adt"
+)
+
+type StateSummary struct {
+	AddrIDs map[addr.Address]abi.ActorID
+	NextID  abi.ActorID
+}
+
+// Checks internal invariants of init state.
+func CheckStateInvariants(st *State, tree *builtin.ActorTree, actorCodes map[string]cid.Cid) (*StateSummary, *builtin.MessageAccumulator) {
```
shrenujbansal, line 20: func signature changed compared to v9

builtin/v10/check.go
```diff
@@ -0,0 +1,359 @@
...
+		case actorCodes[manifest.EvmKey]:
+			var st evm.State
+			if err := tree.Store.Get(tree.Store.Context(), actor.Head, &st); err != nil {
+				fmt.Println("evm invariant error = ", err)
+				return err
+			}
+			msgs := evm.CheckStateInvariants(&st, tree.Store)
+			acc.WithPrefix("evm: ").AddAll(msgs)
+		case actorCodes[manifest.PlaceholderKey]:
+			if err := tree.Store.Get(tree.Store.Context(), actor.Head, &builtin.EmptyState{}); err != nil {
```
arajasek, line 157: I think the correct assertion to make for these three actor types is that the head is always equal to the "empty object" CID. You can see how it's constructed [here](https://github.com/filecoin-project/go-state-types/blob/890a3642f2cbd8fee141f682d17f74db1e8abf7e/builtin/v10/migration/top.go#L354), but I would propose making this a constant in `shared.go` replacing the `EmptyState{}` struct, and referencing that.

builtin/v10/evm/invariants.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"bytes"
+	"github.com/filecoin-project/go-state-types/builtin"
+	"github.com/filecoin-project/go-state-types/builtin/v10/util/adt"
+	"github.com/ipfs/go-cid"
+	"golang.org/x/crypto/sha3"
+	"golang.org/x/xerrors"
+)
+
+// Checks internal invariants of account state.
```
arajasek, line 12: copypasta

builtin/v10/evm/invariants.go
```diff
@@ -0,0 +1,36 @@
...
+func CheckStateInvariants(st *State, store adt.Store) *builtin.MessageAccumulator {
+	acc := &builtin.MessageAccumulator{}
+
+	acc.Require(st.Nonce > 0, "EVM actor state nonce needs to be greater than 0")
```
arajasek, line 16: I didn't realize this was true -- why is it the case?

builtin/v10/evm/invariants.go
```diff
@@ -0,0 +1,36 @@
...
+
+	byteCode, err := getBytecode(st.Bytecode, store)
+	acc.RequireNoError(err, "Unable to retrieve bytecode")
+
+	hasher := sha3.NewLegacyKeccak256()
+	hasher.Write(byteCode)
+	byteCodeHash := hasher.Sum(nil)
+
+	acc.Require(bytes.Equal(byteCodeHash, st.BytecodeHash[:]), "Bytecode hash doesn't match bytecode cid")
```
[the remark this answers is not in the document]
arajasek: Yeah, i think this is a good check to have.

builtin/v10/init/invariants.go
```diff
@@ -0,0 +1,80 @@
...
+		initSummary.AddrIDs[keyAddr] = actorId
+
+		idaddr, err := addr.NewIDAddress(uint64(actorId))
+		acc.RequireNoError(err, "unable to convert actorId %v to id address", actorId)
+		actor, found, err := tree.GetActorV5(idaddr)
+		acc.RequireNoError(err, "unable to retrieve actor with idaddr %v", idaddr)
+		acc.Require(found, "actor not found idaddr %v", idaddr)
+
+		// we expect the address field to be populated for the below actors
+		if (actor.Code == actorCodes[manifest.EthAccountKey] ||
```
arajasek, line 69: I'd like to make two more assertions here:

- If keyAddr is a delegated address, the actor code MUST be one of these 3 types (might be useful having some `isDelegatable` helper, please come up with a better name than I can)
- If an actor in the state tree has an `Address` set, we find a corresponding mapping for it in the init actor (this will need an iteration over the entire actor tree here, or can be a preprocessing step that we do in the top-level iteration before calling the init actor's invariant check).

arajasek (commented):
Thanks for this, this looks pretty good to me! I have a couple suggestions, and look forward to running this on wallayb

builtin/v10/evm/invariants.go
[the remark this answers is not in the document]
shrenujbansal: So this doesn't work at the moment hence the q before I spend more time on it.
I will need to modify the invariants API to have access to the state blockstore since that's where the bytecode resides. I'm not sure if thats fine or we try to avoid it since we only mainly try to look at the actor tree

builtin/v10/init/invariants.go
shrenujbansal: Is #1 dong to be true in the future? From what I understand, the plan is to have delegated addresses for all actors

For #2, this can be added as a cross check between a list built from the top level and the init actor?

arajasek: #1, no it won't, but it will be true for all v10 actors (which is what these invariants are for)
#2 yup, that would work

builtin/v10/check.go
shrenujbansal: done

builtin/v10/init/invariants.go
shrenujbansal: done

builtin/v10/check.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+
+	if err := tree.ForEachV5(func(key address.Address, actor *builtin.ActorV5) error {
+		acc := acc.WithPrefix("%v ", key) // Intentional shadow
+		if key.Protocol() != address.ID {
+			acc.Addf("unexpected address protocol in state tree root: %v", key)
+		}
+		totalFIl = big.Add(totalFIl, actor.Balance)
+
+		if key.Protocol() == address.Delegated {
```
arajasek, line 61: This doesn't work -- see the assertion immediately above this. The only allowed protocol for entries in the state tree are ID addresses.

What you need to do is look at the `actor.Address` field and, if non-nil, assert that it's a Delegated address and put it in this map.

builtin/v10/check.go
```diff
@@ -0,0 +1,367 @@
...
+		}
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+
+	// Check if all delegated addresses are part of init actor
+	for _, addr := range delegatedAddrs {
+		_, found := initSummary.AddrIDs[addr]
+		acc.Require(!found, "delegated address %v not found in init actor map", addr)
```
arajasek, line 181:
```suggestion
		acc.Require(found, "delegated address %v not found in init actor map", addr)
```

Unless I'm an idiot?

shrenujbansal: Nope. I'm the idiot here

builtin/v10/check.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+
+	if err := tree.ForEachV5(func(key address.Address, actor *builtin.ActorV5) error {
+		acc := acc.WithPrefix("%v ", key) // Intentional shadow
+		if key.Protocol() != address.ID {
+			acc.Addf("unexpected address protocol in state tree root: %v", key)
+		}
+		totalFIl = big.Add(totalFIl, actor.Balance)
+
+		if actor.Address != nil && actor.Address.Protocol() == address.Delegated {
```
arajasek, line 61: I'd slightly tweak this: If the address is non-nil, it MUST be a delegated address, so assert that instead of checking-and-skipping.

arajasek (approved):
One nit, but I think this is good.

I would suggest running it once on Wallaby / Hyperspace before merging, because I suspect we might be overlooking something, but we can always correct this if needed.

PR #150, opened 2023-01-13
make gen for eam.CreateExternalReturn

vyzo:
Or else they get no unmarshaller... sigh


arajasek:
Let's just have all 3 as the same type `CreateReturn`, and use those directly.

vyzo:
how about I make them type aliases?

vyzo:
so that we don't have to change anything else elsewhere.

PR #153, opened 2023-01-17
feat: add actors for av11

geoff-vball:


arajasek (approved):
This LGTM, thank you. I'm not sure whether or not this should be landing in `master` -- we're creating this to support hyperspacenet, which needs a separate branch. v10 is pretty stable, but still has some ongoing work that we wound up needing to duplicate for the v11 code in `master`.

TL;DR: I would propose landing this in some separate `ntwk/hyperspace` branch.

PR #162, opened 2023-02-13
evm types construct creator is a 20 byte hash

torchmike:
creator is an eth address, so it can be specified as a 20 byte data array


torchmike:
ajglajsdlfajsdklfj wrong github login

PR #176, opened 2023-03-01
feat: sector migration for nv19

arajasek:
This PR introduces the skeleton of the nv19 migration, including a cached migration that migrates all sectors as needed by https://github.com/filecoin-project/builtin-actors/pull/1229.

Tests will be written in Lotus.

### DIFF
--- a/builtin/v11/migration/miner.go
+++ b/builtin/v11/migration/miner.go
@@ -3,19 +3,77 @@ package migration
 import (
 	"context"
 
+	miner10 "github.com/filecoin-project/go-state-types/builtin/v10/miner"
+	adt10 "github.com/filecoin-project/go-state-types/builtin/v10/util/adt"
+	miner11 "github.com/filecoin-project/go-state-types/builtin/v11/miner"
+	adt11 "github.com/filecoin-project/go-state-types/builtin/v11/util/adt"
+	"github.com/filecoin-project/go-state-types/migration"
+
+	"github.com/filecoin-project/go-address"
+	"github.com/filecoin-project/go-amt-ipld/v4"
 	"golang.org/x/xerrors"
 
-	miner10 "github.com/filecoin-project/go-state-types/builtin/v10/miner"
-	miner11 "github.com/filecoin-project/go-state-types/builtin/v11/miner"
-	"github.com/filecoin-project/go-state-types/migration"
 	"github.com/ipfs/go-cid"
 	cbor "github.com/ipfs/go-ipld-cbor"
+
+	"github.com/filecoin-project/go-state-types/abi"
 )
 
 // The minerMigrator performs the following migrations:
 // - FIP-0061: Updates all miner info PoSt proof types to V1_1 types
+// - #914 fix: Set ActivationEpoch to when the sector was FIRST activated, and PowerBaseEpoch to latest update epoch
+
 type minerMigrator struct {
-	OutCodeCID cid.Cid
+	emptyDeadlineV10  cid.Cid
+	emptyDeadlinesV10 cid.Cid
+	emptyDeadlineV11  cid.Cid
+	emptyDeadlinesV11 cid.Cid
+	OutCodeCID        cid.Cid
+}
+
+func newMinerMigrator(ctx context.Context, store cbor.IpldStore, outCode cid.Cid) (*minerMigrator, error) {
+	ctxStore := adt10.WrapStore(ctx, store)
+
+	edv10, err := miner10.ConstructDeadline(ctxStore)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadline v10: %w", err)
+	}
+
+	edv10cid, err := store.Put(ctx, edv10)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to put empty deadline v10: %w", err)
+	}
+
+	edsv10 := miner10.ConstructDeadlines(edv10cid)
+	edsv10cid, err := store.Put(ctx, edsv10)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadlines v10: %w", err)
+	}
+
+	edv11, err := miner11.ConstructDeadline(ctxStore)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadline v11: %w", err)
+	}
+
+	edv11cid, err := store.Put(ctx, edv11)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to put empty deadline v11: %w", err)
+	}
+
+	edsv11 := miner11.ConstructDeadlines(edv11cid)
+	edsv11cid, err := store.Put(ctx, edsv11)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to construct empty deadlines v11: %w", err)
+
+	}
+
+	return &minerMigrator{
+		emptyDeadlineV10:  edv10cid,
+		emptyDeadlinesV10: edsv10cid,
+		emptyDeadlineV11:  edv11cid,
+		emptyDeadlinesV11: edsv11cid,
+		OutCodeCID:        outCode,
+	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


builtin/v11/miner/miner_state.go
```diff
@@ -170,8 +170,8 @@ type SectorOnChainInfo struct {
 	InitialPledge         abi.TokenAmount // Pledge collected to commit this sector
 	ExpectedDayReward     abi.TokenAmount // Expected one day projection of reward for sector computed at activation time
 	ExpectedStoragePledge abi.TokenAmount // Expected twenty day projection of reward for sector computed at activation time
-	ReplacedSectorAge     abi.ChainEpoch  // Age of sector this sector replaced or zero
-	ReplacedDayReward     abi.TokenAmount // Day reward of sector this sector replace or zero
+	PowerBaseEpoch        abi.ChainEpoch  // Age of sector this sector replaced or zero
+	ReplacedDayReward     abi.TokenAmount // Day reward of sector this sector replaced or zero
```
anorth, line 174: Update the comments too.
Rename ReplacedDayReward if it's subsequently renamed in https://github.com/filecoin-project/builtin-actors/pull/1229.

anorth (changes requested):

builtin/v11/migration/miner.go
```diff
@@ -0,0 +1,317 @@
...
+	if sectorInfo.Activation > powerBaseEpoch {
+		powerBaseEpoch = sectorInfo.Activation
+	}
+
+	return &miner11.SectorOnChainInfo{
+		SectorNumber:          sectorInfo.SectorNumber,
+		SealProof:             sectorInfo.SealProof,
+		SealedCID:             sectorInfo.SealedCID,
+		DealIDs:               sectorInfo.DealIDs,
+		Activation:            sectorInfo.Activation - sectorInfo.ReplacedSectorAge,
```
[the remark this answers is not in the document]
arajasek: You're right -- I think we can simplify PowerBaseEpoch to simply be OldActivation.

builtin/v11/migration/miner.go
```diff
@@ -0,0 +1,317 @@
...
+
+			outDeadlines.Due[i] = outDlCid
+		}
+	}
+
+	return store.Put(ctx, &outDeadlines)
+}
+
+func migrateSectorInfo(sectorInfo miner10.SectorOnChainInfo) *miner11.SectorOnChainInfo {
+	powerBaseEpoch := sectorInfo.ReplacedSectorAge
```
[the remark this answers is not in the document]
arajasek: Thank you!

builtin/v11/migration/top.go
```diff
@@ -0,0 +1,295 @@
...
+	if err != nil {
+		return cid.Undef, xerrors.Errorf("loading state tree: %w", err)
+	}
+	actorsOut, err := builtin.NewTree(adtStore)
+	if err != nil {
+		return cid.Undef, xerrors.Errorf("creating new state tree: %w", err)
+	}
+
+	// load old manifest data
+	systemActor, ok, err := actorsIn.GetActorV5(builtin.SystemActorAddr)
```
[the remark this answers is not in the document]
arajasek: The V5 here is somewhat misleading -- it refers to the State Tree Version 5 (specifically the `ActorV5` struct https://github.com/filecoin-project/go-state-types/blob/393e0b6ffa8839295e6c909b23e57b4a2a099ca5/builtin/actor_tree.go#L23.

builtin/v11/migration/miner.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func migrateSectorInfo(sectorInfo miner10.SectorOnChainInfo) *miner11.SectorOnChainInfo {
+	// For a sector that has not been updated: the Activation is correct and ReplacedSectorAge is zero.
+	// For a sector that has been updated: Activation is the epoch at which it was upgraded, and ReplacedSectorAge is delta since the true activation.
+	//
+	// Thus we want to set:
+	//
+	// Activation := OldActivation - ReplacedSectorAge (a no-op for non-upgraded sectors)
+	// PowerBaseEpoch := Activation (in both upgrade and not-upgraded cases)
```
anorth, line 301:
```suggestion
	// PowerBaseEpoch := OldActivation (in both upgrade and not-upgraded cases)
```

anorth (approved):

anorth:
LGTM, but I recommend having someone more familiar with the migration infrastructure also review.

ZenGround0:
> Tests will be written in Lotus

Have these happened yet?

builtin/v11/migration/system.go
```diff
@@ -0,0 +1,36 @@
+package migration
```
ZenGround0, line 1: Since we are now free away from specs actors versioning needs we should move migration infrastructure here and in top.go into a version independent file.  Then we can have migration specific setting of migrators.

When outer tree structure and system actor manifest state changes we'd have to abstract some of this behind interfaces or copy things like this.

builtin/v11/migration/miner.go
```diff
@@ -0,0 +1,323 @@
...
+				return cid.Undef, err
+			}
+
+			outSectorsSnapshotCid, err := cache.Load(migration.SectorsAmtKey(inDeadline.SectorsSnapshot), func() (cid.Cid, error) {
+				inSectorsSnapshot, err := adt10.AsArray(store, inDeadline.SectorsSnapshot, miner10.SectorsAmtBitwidth)
+				if err != nil {
+					return cid.Undef, err
+				}
+
+				outSectorsSnapshot, err := migrateSectorsFromScratch(ctx, store, inSectorsSnapshot)
```
ZenGround0, line 256: Time to bring this back up: I don't understand why we go to the trouble of AMT diffing for sectors tree but not for sector snapshots.  Have we confirmed that not much time is spent here?  I think we should confirm that and if we actually do spend a decent fraction of time here design something nice.

I think we can diff against the latest sector tree assuming the cache for this miner address is populated.  Because this means that snapshot has changed since pre migration which means that the snapshot tree will be very close to the currently cached sectors AMT.

It will likely also make the premigration a lot faster if we do sectors tree first and then diff all of these against it.  Though this depends on AMT diffing performance which maybe I'm overestimating.

I'm suspicious we can do all of this with 0 caching overhead by just referencing already cached values.  

Why do I think its worth it now? If we do this now we can either find there's no big speed up and drop forever or find there is a nice speedup and finish this now and reuse on all subsequent sector info migrations (like we are doing with the nice diffing code on sectors AMT)

builtin/v11/migration/miner.go
```diff
@@ -0,0 +1,323 @@
...
+	// For a sector that has not been updated: the Activation is correct and ReplacedSectorAge is zero.
+	// For a sector that has been updated: Activation is the epoch at which it was upgraded, and ReplacedSectorAge is delta since the true activation.
+	//
+	// Thus we want to set:
+	//
+	// Activation := OldActivation - ReplacedSectorAge (a no-op for non-upgraded sectors)
+	// PowerBaseEpoch := Activation (in both upgrade and not-upgraded cases)
+
+	powerBaseEpoch := sectorInfo.Activation
+	activationEpoch := sectorInfo.Activation - sectorInfo.ReplacedSectorAge
```
ZenGround0, line 304: Curious here -- do we have any really old school CC upgrade sectors with a non zero Replaced sector age but a true activation time set at Activation ?  Its worth raising a flag in a practice migration to see if this is something we need to handle.  I think we are safe if we only make this change if SectorKey != nil

builtin/v11/migration/miner.go
ZenGround0: I tried this out and empirically we performed *worse* on the premigration.  Unclear why unless AMT diffing is generally inefficient.  Its worth continuing to understand this in the future but for now leaving this is the way to go.

ZenGround0 (approved):

PR #181, opened 2023-03-30
refactor: verified registry: don't error if map entry isn't found

arajasek:
Also chain errors where missing.

### DIFF
--- a/builtin/v11/verifreg/verified_registry_state.go
+++ b/builtin/v11/verifreg/verified_registry_state.go
@@ -72,9 +72,13 @@ func (st *State) FindAllocation(store adt.Store, clientIdAddr address.Address, a
 		return nil, false, xerrors.Errorf("can only look up ID addresses")
 	}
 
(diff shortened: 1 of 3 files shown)


builtin/v11/verifreg/verified_registry_state.go
```diff
@@ -72,9 +72,13 @@ func (st *State) FindAllocation(store adt.Store, clientIdAddr address.Address, a
...
 	}
 
-	innerHamtCid, err := getInnerHamtCid(store, abi.IdAddrKey(clientIdAddr), st.Allocations)
+	innerHamtCid, ok, err := getInnerHamtCid(store, abi.IdAddrKey(clientIdAddr), st.Allocations)
 	if err != nil {
-		return nil, false, err
+		return nil, false, xerrors.Errorf("failed to get inner hamt cid: %w", err)
+	}
+
+	if !ok {
```
ZenGround0, line 80: nit: maybe a bit clearer to explicitly return false when !ok

ZenGround0 (approved):

PR #182, opened 2023-04-05
feat: Implement FIP-0061

arajasek:


builtin/v11/migration/top.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	// Maps prior version code CIDs to migration functions.
+	migrations := make(map[cid.Cid]migration.ActorMigration)
+	// Set of prior version code CIDs for actors to defer during iteration, for explicit migration afterwards.
+	deferredCodeIDs := make(map[cid.Cid]struct{})
+
+	miner10Cid := cid.Undef
+	for _, oldEntry := range oldManifestData.Entries {
+		if oldEntry.Name == manifest.MinerKey {
```
ZenGround0, line 71: nit: No correctness problem since you overwrite below, but to be consistent and clear I think you should do the same thing with oldEntry.Name == manifest.SystemKey in another if case branch OR remove this if statement and just overwrite the miner migration the way you overwrite the system migration.

ZenGround0 (approved):
Im assuming all the migration/runner and migration/job stuff is the same as the already reviewed SDM PR.  LGTM

Stebalien:
Note: the FFI currently depends on this PR instead of a released state-types. When this PR is merged, please update the FFI.

ZenGround0 (approved):

PR #185, opened 2023-04-21
fix: Add missing post proof

strahe:


arajasek (approved):
Thanks for the catch!

PR #188, opened 2023-04-24
Remove incorrect invariants

ZenGround0:
I'm still not 100% sure its right to remove these.  They are definitely incorrect but invariant breaks of these forms are somewhat useful to see in the logs as they indicate particular edge cases that can be confirmed to not be part of the problem.  I think it's probably better to remove them for cleanliness but ideally we'd try to find a way to assert something about the edge cases.  Right now that looks either impossible or difficult hence this PR.

### DIFF
--- a/builtin/v11/check.go
+++ b/builtin/v11/check.go
@@ -355,7 +355,9 @@ func CheckMarketAgainstVerifreg(acc *builtin.MessageAccumulator, verifregSummary
 	for allocationId, dealId := range marketSummary.AllocIdToDealId {
 		alloc, found := verifregSummary.Allocations[allocationId]
 		acc.Require(found, "allocation %d not found for pending deal %d", allocationId, dealId)
-
+		if !found {
+			continue
+		}
 		info, found := marketSummary.Deals[dealId]
 		acc.Require(found, "internal invariant error invalid market state references missing deal %d", dealId)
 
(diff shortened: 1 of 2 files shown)


builtin/v11/check.go
```diff
@@ -327,11 +327,8 @@ func CheckVerifregAgainstMiners(acc *builtin.MessageAccumulator, verifregSummary
 		maddr, err := address.NewIDAddress(uint64(claim.Provider))
 		acc.RequireNoError(err, "error creating ID address: %v", err)
 
-		minerSummary, ok := minerSummaries[maddr]
+		_, ok := minerSummaries[maddr]
```
ZenGround0, line 330: Claims still exist after sector termination which is where all of the failures observed on mainnet traced back to.

More concerning is that claims still exist after expiration until cleanup happens.  Once simple qap FIL+ sectors start expiring this message is going to become very spammy.

ZenGround0: To better assert this we could change miner summaries to find terminated sectors which has its own problems as termination information can be lost today.  We could also look at the writeonly sectors bitfield for a weaker condition but that could hide problems as no sector need to be associated with it.

builtin/v11/init/invariants.go
```diff
@@ -63,8 +63,9 @@ func CheckStateInvariants(st *State, tree *builtin.ActorTree, actorCodes map[str
 		acc.RequireNoError(err, "unable to convert actorId %v to id address", actorId)
 		actor, found, err := tree.GetActorV5(idaddr)
 		acc.RequireNoError(err, "unable to retrieve actor with idaddr %v", idaddr)
-		acc.Require(found, "actor not found idaddr %v", idaddr)
-
+		if !found {
+			return nil // this can happen if actor self destructs as init is not informed
```
ZenGround0, line 67: On mainnet I've only seen Paych actors but I believe EVM actors can also self destruct.

It is hard to assert anything about these self destructed actors (like code cid) because the state itself has lost all information about the actor after self destruct.

arajasek (approved):

PR #191, opened 2023-06-27
PorepID / ReplicaID utils on RegisteredSealProof

magik6k:
With new, lower level proof APIs in Go, these will be quite useful


magik6k:
(need to confirm the outputs are correct)

abi/actor.go
```diff
@@ -12,6 +16,18 @@ func (e ActorID) String() string {
 	return strconv.FormatInt(int64(e), 10)
 }
 
+// ProverID returns a 32 byte proverID used when computing ReplicaID
+func (e ActorID) ProverID() ([32]byte, error) {
```
arajasek, line 20: Instead of defining this on the `ActorID` type, can we make it a `ProverIDFromMinerIDForPoRepReplicaID` (or some less ridiculous name) helper that we put in `sector.go` or something?

abi/sector.go
```diff
@@ -299,6 +300,98 @@ func (p RegisteredSealProof) RegisteredUpdateProof() (RegisteredUpdateProof, err
...
+	RegisteredSealProof_StackedDrg64GiBV1:  4,
+
+	RegisteredSealProof_StackedDrg2KiBV1_1:   5,
+	RegisteredSealProof_StackedDrg8MiBV1_1:   6,
+	RegisteredSealProof_StackedDrg512MiBV1_1: 7,
+	RegisteredSealProof_StackedDrg32GiBV1_1:  8,
+	RegisteredSealProof_StackedDrg64GiBV1_1:  9,
+}
+
+func (p RegisteredSealProof) porepNonce() uint64 {
```
arajasek, line 319: Just make it a private constant for now?

arajasek (commented):
> With new, lower level proof APIs in Go, these will be quite useful

Can you elaborate on this? I don't have any opposition to this PR, I just don't know what this means.

abi/sector.go
```diff
@@ -299,6 +300,98 @@ func (p RegisteredSealProof) RegisteredUpdateProof() (RegisteredUpdateProof, err
...
+	// https://github.com/filecoin-project/rust-filecoin-proofs-api/blob/9b580c2791f028b7ce3aaef0cf9d68956c50170d/src/registry.rs#L166
+	return 0
+}
+
+// PoRepID produces the porep_id for this RegisteredSealProof. Mainly used for
+// computing replica_id.
+func (p RegisteredSealProof) PoRepID() ([32]byte, error) {
+	// https://github.com/filecoin-project/rust-filecoin-proofs-api/blob/9b580c2791f028b7ce3aaef0cf9d68956c50170d/src/registry.rs#L174
```
shrenujbansal, line 327: This line no might not be valid post any change to this file. Maybe add the func name in the comment?

abi/sector.go
```diff
@@ -299,6 +300,98 @@ func (p RegisteredSealProof) RegisteredUpdateProof() (RegisteredUpdateProof, err
...
+	copy(id[0:8], proofIdBytes)
+	copy(id[8:16], nonceBytes)
+
+	return id, nil
+}
+
+// ReplicaId produces the replica_id for this RegisteredSealProof. This is used
+// as the main input for computing SDR
+func (p RegisteredSealProof) ReplicaId(prover ActorID, sector SectorNumber, ticket []byte, commd []byte) ([32]byte, error) {
+	// https://github.com/filecoin-project/rust-fil-proofs/blob/5b46d4ac88e19003416bb110e2b2871523cc2892/storage-proofs-porep/src/stacked/vanilla/params.rs#L758-L775
```
shrenujbansal, line 353: Same as above

abi/sector.go
magik6k: This links to a commit, not the current branch

abi/sector.go
magik6k: I wanted to match the rust code so that when this needs to change there's easy mapping between methods

abi/actor.go
magik6k: Why? That method would be taking the ActorID anyways, except now would be harder to discover, and have a longer name.

arajasek: Because `ActorID` doesn't have anything to do with PoRep / Replicas / proofs. It's not miner-specific, just the key of the state tree, so defining some method on it for the sake of unrelated code that 99% of its users won't actually care about seems bad.

abi/sector.go
arajasek: That seems reasonable

shrenujbansal (approved):

PR #199, opened 2023-08-02
Migrate deadlines using diff

ZenGround0:
This is a speculative optimization.  The idea is that we might be doing many loads migrations and writes of effecitively the same data if there are differences in the snapshotted sectors amts inside of miner deadlines.  This change does sector info migrations on each sector info exactly once.


snissn:
Re-opening this because our tests have confirmed its correctness https://github.com/filecoin-project/lotus/pull/11149

snissn:
Re-opening this because our tests have confirmed its correctness https://github.com/filecoin-project/lotus/pull/11149

snissn (approved):
this code is much faster and is verified to be correct for our migrations

PR #202, opened 2023-08-23
Mikers/sector deal ids index

snissn:
see related issue here https://github.com/filecoin-project/lotus/issues/11160

### DIFF
--- a/builtin/v12/market/cbor_gen.go
+++ b/builtin/v12/market/cbor_gen.go
@@ -20,7 +20,7 @@ var _ = cid.Undef
 var _ = math.E
 var _ = sort.Sort
 
-var lengthBufState = []byte{140}
+var lengthBufState = []byte{141}
 
 func (t *State) MarshalCBOR(w io.Writer) error {
 	if t == nil {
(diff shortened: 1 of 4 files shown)


builtin/v12/market/cbor_gen.go
```diff
@@ -1784,6 +1802,156 @@ func (t *OnMinerSectorsTerminateParams) UnmarshalCBOR(r io.Reader) (err error) {
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err == io.EOF {
+			err = io.ErrUnexpectedEOF
+		}
+	}()
+
+	if maj != cbg.MajArray {
```
ZenGround0, line 1932: This isn't quite the right serialization for a cid.  By default cbor gen is going to see a struct and serialize it as an array.  If you want a HamtCid type that will cborgen correctly by default you could make an alias type.  

While I sympathize with the instinct behind the special type given that our untyped raw pointers are not great I would probably avoid breaking our not-so-great convention.  I would avoid specially typing HamtCid altogether and just use the not-so-great convention of a plain cid.

builtin/v12/migration/top.go
```diff
@@ -66,18 +68,30 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
 	deferredCodeIDs := make(map[cid.Cid]struct{})
 
 	miner11Cid := cid.Undef
+	market11Cid := cid.Undef
 
 	for _, oldEntry := range oldManifestData.Entries {
 		if oldEntry.Name == manifest.MinerKey {
 			miner11Cid = oldEntry.Code
 		}
+		if oldEntry.Name == manifest.MinerKey {
```
snissn, line 77: MarketKey

snissn:
out of date - code is in #210

PR #207, opened 2023-08-23
chore: nv21 dev reverts

magik6k:
Part of https://github.com/filecoin-project/go-state-types/issues/203

This PR temporarily moves nv21 changes off master (nv21 changes are in `feat/nv21`)

Issue #203, opened 2023-08-23
Branch cleanup, move nv21 work to a staging branch


magik6k:
* [x] Setup feat/nv21 from current master
  * https://github.com/filecoin-project/go-state-types/tree/feat/nv21
* [x] Start a revert branch
* [x] Reset to ff2ed169ff566458f2acd8b135d62e8ca27e7d0c
* [x] Cherry-pick #191
* [x] Open a PR reapplying feat/nv21 changes to the revert branch (which will become master when revert branch is merged) - https://github.com/filecoin-project/go-state-types/pull/208
* [ ] Merge the revert branch into master

arajasek:
Hang on, I don't think we're aligned here, this seems unnecessary and bad to me. I'd like us to leave _as much_ nv21 stuff as possible in `master`. Anything that is non-breaking, like the new network version, the new actor types, the migration, etc. can stay there. I think we _only_ need to stage changes that can't currently be pulled into Lotus `master` because it would break things. I'm not sure what those specific changes are, but I'd like us to only revert those changes into a dedicated staging branch like `feat/synth-porep` or something. For all other purposes `master` can be the nv21 branch.

This is what we've historically done, and it's worked pretty well. From my perspective, this might only be a problem if we ever have to urgently ship a _different_ network upgrade, and so might wind up having to revert changes. Even in that case, though, I think we'd simply force reset to the "clean" commit that adds the new types identical to the previous version (we don't have an example for v12, but [here's the v11 one](https://github.com/filecoin-project/go-state-types/commits/master)), and then make whatever new changes we need.

magik6k:
After some back-and-forth it looks like we don't really need to do this, and it will be cleaner to just merge the current lotus feat/nv21 into lotus master, at which point anything else can start pulling in latest GST

PR #208, opened 2023-08-23
Support for nv21

magik6k:
Part of https://github.com/filecoin-project/go-state-types/issues/203

For now based on https://github.com/filecoin-project/go-state-types/pull/207, but the base should switch to master once that PR is merged.

Only contains code from commits which were reverted in #207

Code as of f65f7b86b should be identical to a154da5; Check with:

```
$ git checkout f65f7b86b
$ git diff a154da5
[empty]
```


magik6k:
Not really needed - see https://github.com/filecoin-project/go-state-types/issues/203#issuecomment-1690144648

PR #210, opened 2023-08-29
feat: migration: sector deal ids index cached

snissn:
- Add migration cache to newMinerMigrator constructor.
- Implement sector index caching in migrateSectorsWithCache.
- Optimize sector migration by leveraging diffs in cached AMT.
- Manage sector-to-deal mappings using a HAMT structure.


builtin/v12/migration/miner.go
```diff
@@ -213,6 +241,9 @@ func (m minerMigrator) migrateSectorsWithCache(ctx context.Context, store adt11.
 			return cid.Undef, xerrors.Errorf("failed to write inkey to cache: %w", err)
 		}
 
+		if err = cache.Write(migration.SectorIndexPrevSectorsOutKey(minerAddr), sectorToDealIdHamtCid); err != nil {
```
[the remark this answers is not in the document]
snissn: resolved

builtin/v12/migration/miner.go
```diff
@@ -28,10 +33,12 @@ type minerMigrator struct {
 	emptyDeadlinesV11 cid.Cid
 	emptyDeadlineV12  cid.Cid
 	emptyDeadlinesV12 cid.Cid
+	sectorDeals       *builtin.ActorTree
 	OutCodeCID        cid.Cid
+	hamtLock          *sync.Mutex
```
ZenGround0, line 38: nit: since there are a lot of hamts in the miner actor give this a slightly more descriptive name i.e. sectorDealsLock.  Also consider renaming sectorDeals to something that better communicates its global nature (i.e. marketSectorDeals, globalSectorDeals, allSectorDeals..)

builtin/v12/migration/miner.go
```diff
@@ -28,10 +33,12 @@ type minerMigrator struct {
 	emptyDeadlinesV11 cid.Cid
 	emptyDeadlineV12  cid.Cid
 	emptyDeadlinesV12 cid.Cid
+	sectorDeals       *builtin.ActorTree
```
ZenGround0, line 36: nit: use adt.Map directly instead of the extraneous ActorTree wrapping type

builtin/v12/migration/miner.go
ZenGround0: Also Im quite confused how this migration is working at all since hamtLock never appears to be initialized.  You should be getting a panic when you call this: https://go.dev/play/p/_7PCDPsWrac

builtin/v12/migration/top.go
```diff
@@ -95,25 +109,85 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
+	}
+
 	actorsOut, err := migration.RunMigration(ctx, cfg, cache, store, log, actorsIn, migrations)
 	if err != nil {
 		return cid.Undef, xerrors.Errorf("failed to run migration: %w", err)
 	}
 
-	outCid, err := actorsOut.Flush()
+	// Retrieve the sector deal IDs from the minerMigrator. These IDs are crucial for the migration of market actor state.
+	sectorDealIDs, err := minerMigrator.sectorDeals.Map.Root()
```
ZenGround0, line 143: This approach relies on HAMT root being deterministic independent of insertion order.  Looking at the code this seems plausible but I want to confirm this is an expected property before relying on this

cc @Stebalien @rvagg

builtin/v12/migration/miner.go
```diff
@@ -405,3 +499,97 @@ func migrateSectorInfo(sectorInfo miner11.SectorOnChainInfo) *miner12.SectorOnCh
...
+	}); err != nil {
+		return nil, err
+	}
+
+	return outArray, err
+}
+
+func addSectorNumberToDealIdHAMT(xap *builtin.ActorTree, sectorInfo miner11.SectorOnChainInfo, store adt11.Store) error {
+	//sectorInfo.DealIDs,sectorInfo.SectorNumber
+	// - todo confirm types beint stored is correct
```
ZenGround0, line 566: Type looks good, an integration test including messages that read and write this index after migration construction will fully confirm.

builtin/v12/migration/miner.go
```diff
@@ -405,3 +499,97 @@ func migrateSectorInfo(sectorInfo miner11.SectorOnChainInfo) *miner12.SectorOnCh
...
+	err := xap.Map.Delete(abi.IntKey(int64(SectorNumber)))
+	if err != nil {
+		return xerrors.Errorf("failed to delete sector from sectorToDealIdHamt index: %w", err)
+	}
+	return nil
+}
+
+func (m minerMigrator) addSectorToDealIDHamtToSectorDeals(hamtCid cid.Cid, minerAddr address.Address) error {
+	m.hamtLock.Lock()
+	// Lock so only one goroutine at a time can access the map c.v.
```
ZenGround0, line 586: nit: you can probably remove this as its fairly clear from context

builtin/v12/migration/miner.go
```diff
@@ -405,3 +499,97 @@ func migrateSectorInfo(sectorInfo miner11.SectorOnChainInfo) *miner12.SectorOnCh
...
+
+	return outArray, err
+}
+
+func addSectorNumberToDealIdHAMT(xap *builtin.ActorTree, sectorInfo miner11.SectorOnChainInfo, store adt11.Store) error {
+	//sectorInfo.DealIDs,sectorInfo.SectorNumber
+	// - todo confirm types beint stored is correct
+
+	cborDealIDs := market.SectorDealIDs{DealIDs: sectorInfo.DealIDs}
+	err := xap.Map.Put(abi.IntKey(int64(sectorInfo.SectorNumber)), &cborDealIDs)
```
ZenGround0, line 569: nit: `xap.Map.Put(abi.IntKey(int64(sectorInfo.SectorNumber)), &market.SectorDealIDs{DealIDs: sectorInfo.DealIDs})` since IIRC go lets you get away with construction and serialization in one line.

builtin/v12/migration/miner.go
```diff
@@ -405,3 +499,97 @@ func migrateSectorInfo(sectorInfo miner11.SectorOnChainInfo) *miner12.SectorOnCh
...
+	}
+	return nil
+}
+
+func (m minerMigrator) addSectorToDealIDHamtToSectorDeals(hamtCid cid.Cid, minerAddr address.Address) error {
+	m.hamtLock.Lock()
+	// Lock so only one goroutine at a time can access the map c.v.
+	defer m.hamtLock.Unlock()
+
+	err := m.sectorDeals.Map.Put(abi.IdAddrKey(minerAddr), cbg.CborCid(hamtCid))
```
ZenGround0, line 589: The problem with relying on the cached value of sectorDeals is that a miner actor might have recently cleaned up all sectors with deal ids.  This miner addr's HAMT will not be overwritten here so it will be written into market actor state. These ids will never be cleaned up by the system because they have already expired.  

I would suggest experimenting with no caching at the top level structure at all to address this easily.  I don't expect duplicate writes of this HAMT to the blockstore to be taking that much time.

If you want to keep the caching structure then you'll need to track all keys which were not updated by keeping updated keys in a map and then comparing to the cached value.  At this point I would probably just make the minerMigrator's sectorDeals struct a golang map, write to it one time and then do a for each of the cached top level HAMT at the end, deleting any keys that don't show up in this map and overwriting any that are different.

ZenGround0 (commented):
Unfortunately I missed a large and important piece of this migration when we first went over the specification of what we want.

This migration creates an index over all deal ids referenced by sector infos.  However half of all sector infos are dead and their deal ids are therefore also dead / about to be dead.  We want to only add deal ids that are active in the market actor to the index, otherwise they will never be cleaned up.

I think the easiest way to do this is create a set of active deal ids by reading the v11 market actor state before the miner migrations, passing this into the miner migrator and then only adding deal ids that are in this set.

builtin/v12/migration/top.go
ZenGround0: After checking in with others and realizing that `RunMigration` relies on this being the case when setting results in `actorsOut` to make the top level actors HAMT I am convinced this is not a problem.

builtin/v12/migration/miner.go
```diff
@@ -224,65 +271,112 @@ delta = prevInRoot - inRoot
...
+	ret, err := sectorToDealIdHamt.Map.Root()
+	if err != nil {
+		return cid.Undef, cid.Undef, xerrors.Errorf("failed to get root of sectorToDealIdHamt: %w", err)
+	}
+	return prevOutSectorsRoot, ret, nil
+
 }
 
-func migrateSectorsFromScratch(ctx context.Context, store adt11.Store, inArray *adt11.Array) (*adt12.Array, error) {
+// return cid of the HAMT for this address and add to the cache outside this fn
```
ZenGround0, line 352: This comment is confusing outside the context of this PR, please either drop or write a function comment describing the function and its inputs / outputs whollistically .

builtin/v12/migration/miner.go
```diff
@@ -224,65 +271,112 @@ delta = prevInRoot - inRoot
...
+func (m minerMigrator) migrateSectorsFromScratch(ctx context.Context, store adt11.Store, minerAddr address.Address, inArray *adt11.Array) (*adt12.Array, cid.Cid, error) {
 	outArray, err := adt12.MakeEmptyArray(store, miner12.SectorsAmtBitwidth)
 	if err != nil {
-		return nil, xerrors.Errorf("failed to construct new sectors array: %w", err)
+		return nil, cid.Undef, xerrors.Errorf("failed to construct new sectors array: %w", err)
+	}
+
+	sectorToDealIdHamt, err := builtin.NewTree(store)
```
ZenGround0, line 359: nit: use adt.Map directly

builtin/v12/migration/miner.go
```diff
@@ -371,7 +465,7 @@ func (m minerMigrator) migrateDeadlines(ctx context.Context, store adt11.Store,
 	return store.Put(ctx, &outDeadlines)
 }
 
-func migrateSectorInfo(sectorInfo miner11.SectorOnChainInfo) *miner12.SectorOnChainInfo {
+func migrateSectorInfo(sectorInfo miner11.SectorOnChainInfo, store adt11.Store) *miner12.SectorOnChainInfo {
```
ZenGround0, line 468: I think you can remove this argument

builtin/v12/migration/miner.go
```diff
@@ -224,65 +271,112 @@ delta = prevInRoot - inRoot
...
+			// add sector to the HAMT
+			err = addSectorNumberToDealIdHAMT(sectorToDealIdHamt, *info, store)
+			if err != nil {
+				return cid.Undef, cid.Undef, xerrors.Errorf("failed to add sector %d to HAMT: %w", sectorNo, err)
 			}
 		}
 	}
 
-	return prevOutSectors.Root()
+	//return the hAmt
```
ZenGround0, line 337: nit: Either remove or be more specific that the hamt is the sector id => deal id mapping

builtin/v12/migration/miner.go
```diff
@@ -224,65 +271,112 @@ delta = prevInRoot - inRoot
 
 merge( prevOutRoot, Migrated(delta) )
 */
-func migrateSectorsWithDiff(ctx context.Context, store adt11.Store, inRoot cid.Cid, prevInRoot cid.Cid, prevOutRoot cid.Cid) (cid.Cid, error) {
-	// we have previous work, but the AMT has changed -- diff them
+
+// todo modify hamt based on diff
+func (m minerMigrator) migrateSectorsWithDiff(ctx context.Context, store adt11.Store, minerAddr address.Address, inRoot cid.Cid, prevInRoot cid.Cid, prevOutRoot cid.Cid, prevSectorIndexRoot cid.Cid) (cid.Cid, cid.Cid, error) {
```
ZenGround0, line 276: Not a requirement for merging but I suspect this is going to come out overall cleaner if you share code with the migrateDeadlineSectorsWithDiff function.  I would probably keep two separate function calls with the same interface you have and then call into a shared internal function which takes a boolean flag for operating over the sectorNumber => DealID index.

builtin/v12/migration/miner.go
```diff
@@ -185,14 +220,23 @@ func migrateSectorsWithCache(ctx context.Context, store adt11.Store, cache migra
...
+			outArray, outDealHamtCid, err := m.migrateSectorsFromScratch(ctx, store, minerAddr, inArray)
 			if err != nil {
 				return cid.Undef, xerrors.Errorf("failed to migrate sectors from scratch: %w", err)
 			}
 			outRoot, err = outArray.Root()
 			if err != nil {
 				return cid.Undef, xerrors.Errorf("error writing new sectors AMT: %w", err)
 			}
+
+			sectorToDealIdHamtCid = outDealHamtCid
```
ZenGround0, line 233: nit: I would just use one variable name in both if branches to reduce noise.

builtin/v12/migration/miner.go
ZenGround0: Same thing applies to the separate migrateDeadlineSectorsFromScratch function

builtin/v12/migration/top.go
rvagg: :thumbsup: that's one of the fundamental properties that makes us so keen on HAMTs for content addressed data structures. If it ends up not being true then we have a bug!

anorth:
See #215, which should be integrated into this change.

migration/util.go
```diff
@@ -133,6 +133,14 @@ func ExpirationsAmtKey(expirationsAmt cid.Cid) string {
 	return "partitionsAmt-" + expirationsAmtKey
 }
 
+func SectorIndexHamtKey() string {
+	return "sectorIndexHamtKey"
+}
+
+func SectorIndexPrevSectorsOutKey(m address.Address) string {
```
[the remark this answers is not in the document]
snissn: #98270db

migration/util.go
```diff
@@ -133,6 +133,14 @@ func ExpirationsAmtKey(expirationsAmt cid.Cid) string {
 	return "partitionsAmt-" + expirationsAmtKey
 }
 
+func SectorIndexHamtKey() string {
```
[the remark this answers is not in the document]
snissn: #da0545e

builtin/v12/migration/miner.go
snissn: #a697049

builtin/v12/migration/miner.go
snissn: done

builtin/v12/migration/miner.go
snissn: I didn't understand this one -- when i code it what should be the normal way:

```
mikers@mikers-B560-DS3H-AC-Y1:~/dev/filecoin-project/go-state-types$ git diff 
diff --git a/builtin/v12/migration/miner.go b/builtin/v12/migration/miner.go
index 4abacc4..f3216b0 100644
--- a/builtin/v12/migration/miner.go
+++ b/builtin/v12/migration/miner.go
@@ -221,7 +221,7 @@ func (m minerMigrator) migrateSectorsWithCache(ctx context.Context, store adt11.
                                return cid.Undef, xerrors.Errorf("failed to read sectors array: %w", err)
                        }
 
-                       outArray, outDealHamtCid, err := m.migrateSectorsFromScratch(ctx, store, minerAddr, inArray)
+                       outArray, sectorToDealIdHamtCid, err := m.migrateSectorsFromScratch(ctx, store, minerAddr, inArray)
                        if err != nil {
                                return cid.Undef, xerrors.Errorf("failed to migrate sectors from scratch: %w", err)
                        }
@@ -229,8 +229,6 @@ func (m minerMigrator) migrateSectorsWithCache(ctx context.Context, store adt11.
                        if err != nil {
                                return cid.Undef, xerrors.Errorf("error writing new sectors AMT: %w", err)
                        }
-
-                       sectorToDealIdHamtCid = outDealHamtCid
                }
 
                //add to new address sector deal id hamt index
```

i get this compiler error:

builtin/v12/migration/miner.go:224:14: sectorToDealIdHamtCid declared but not used

as if sectorToDealIdHamtCid is a new variable just within the if statement scope.. i found it works if i use a tmp variable but i don't see what the compilers issue is

builtin/v12/migration/miner.go
snissn: #ffbf314 done

builtin/v12/migration/miner.go
snissn: //return the sector to deal id Hamt

builtin/v12/migration/miner.go
snissn: thanks, cool! the tests will be in lotus itests

snissn:
Thanks @ZenGround0 I went through and resolved the easy notes, looks like remaining i have to:

1. not construct the parent HAMT during the caching process and only at the end primarily because we will potentially keep data on miners that have been deleted between running the pre migration and the actual migration
2. i'm using builtin.NewTree which is a wrapper around adt.Map and should just use adt.Map directly
3. put market migration in it's own file to run after the migration ( might be need to get help w subtleties in this one)

ZenGround0:
The other thing is probably the most annoying change to filter out for inactive deal ids.  Probably got lost among the rest since I left it in the review comment itself: https://github.com/filecoin-project/go-state-types/pull/210#pullrequestreview-1608361440

builtin/v12/migration/miner.go
snissn: it's a six vs half a dozen situation, going to leave it as is

builtin/v12/migration/miner.go
snissn: done! :)

builtin/v12/migration/miner.go
snissn: > The problem with relying on the cached value of sectorDeals is that a miner actor might have recently cleaned up all sectors with deal ids. This miner addr's HAMT will not be overwritten here so it will be written into market actor state. 

incorrect, because of the diff that is involved, so current code is correct

builtin/v12/migration/top.go
```diff
@@ -95,25 +109,85 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
+	newMarketStateCid, err := store.Put(ctx, &newMarketState)
 	if err != nil {
-		return cid.Undef, xerrors.Errorf("failed to flush actorsOut: %w", err)
+		return cid.Undef, xerrors.Errorf("failed to put new state: %w", err)
+	}
+
+	// Update the market actor in the state tree with the newly fetched sector deal IDs.
+	// This ensures the market actor's state reflects the most recent sector deals.
+	if err = actorsOut.SetActorV5(builtin.StorageMarketActorAddr, &builtin.ActorV5{
+		Code:       market11Cid,
```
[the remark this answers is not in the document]
snissn: ty! done :D

builtin/v12/migration/top.go
```diff
@@ -95,25 +109,85 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
 		return cid.Undef, xerrors.Errorf("failed to create miner migrator: %w", err)
 	}
-	migrations[miner11Cid] = migration.CachedMigration(cache, *mm)
+	migrations[miner11Cid] = migration.CachedMigration(cache, *minerMigrator)
 
 	if len(migrations)+len(deferredCodeIDs) != len(oldManifestData.Entries) {
 		return cid.Undef, xerrors.Errorf("incomplete migration specification with %d code CIDs, need %d", len(migrations), len(oldManifestData.Entries))
 	}
 
+	// Load the state of the market actor from v11 for migration purposes.
```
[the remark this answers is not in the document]
snissn: keeping in top.go

builtin/v12/migration/top.go
```diff
@@ -66,18 +68,30 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
 	deferredCodeIDs := make(map[cid.Cid]struct{})
```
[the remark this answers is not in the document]
snissn: ty!!!

builtin/v12/migration/miner.go
```diff
@@ -224,65 +272,111 @@ delta = prevInRoot - inRoot
...
 	prevOutSectors, err := miner12.LoadSectors(store, prevOutRoot)
 	if err != nil {
-		return cid.Undef, xerrors.Errorf("failed to load prevOutSectors: %w", err)
+		return cid.Undef, cid.Undef, xerrors.Errorf("failed to load prevOutSectors: %w", err)
+	}
+
+	// load previous HAMT sector index for this specific minerAddr
+	sectorToDealIdHamt, err := builtin.LoadTree(store, prevSectorIndexRoot)
+	if err != nil {
+		return cid.Undef, cid.Undef, xerrors.Errorf("creating new state tree: %w", err)
```
arajasek, line 297: bad error message

builtin/v12/migration/miner.go
```diff
@@ -405,3 +499,92 @@ func migrateSectorInfo(sectorInfo miner11.SectorOnChainInfo) *miner12.SectorOnCh
...
+	if err = inArray.ForEach(&sectorInfo, func(k int64) error {
+		return outArray.Set(uint64(k), migrateSectorInfo(sectorInfo))
+	}); err != nil {
+		return nil, err
+	}
+
+	return outArray, err
+}
+
+func addSectorNumberToDealIdHAMT(xap *builtin.ActorTree, sectorInfo miner11.SectorOnChainInfo, store adt11.Store) error {
```
arajasek, line 564: what on earth is an xap?

builtin/v12/migration/miner.go
```diff
@@ -224,65 +272,111 @@ delta = prevInRoot - inRoot
...
 	}
-
 	prevOutSectors, err := miner12.LoadSectors(store, prevOutRoot)
 	if err != nil {
-		return cid.Undef, xerrors.Errorf("failed to load prevOutSectors: %w", err)
+		return cid.Undef, cid.Undef, xerrors.Errorf("failed to load prevOutSectors: %w", err)
+	}
+
+	// load previous HAMT sector index for this specific minerAddr
+	sectorToDealIdHamt, err := builtin.LoadTree(store, prevSectorIndexRoot)
```
arajasek, line 295: it's probably safe, but a little wrong to use `LoadTree` here -- that method returns an `ActorTree`, which at least on paper is supposed to be only for "state trees", and not just any HAMT

builtin/v12/migration/miner.go
```diff
@@ -224,65 +272,111 @@ delta = prevInRoot - inRoot
...
 		return outArray.Set(uint64(k), migrateSectorInfo(sectorInfo))
 	}); err != nil {
-		return nil, err
+		return nil, cid.Undef, err
 	}
 
-	return outArray, err
+	sectorToDealIdHamtCid, err := sectorToDealIdHamt.Map.Root()
+	if err != nil {
+		return nil, cid.Undef, err
```
arajasek, line 376: chain errors pls

builtin/v12/migration/miner.go
```diff
@@ -120,7 +149,7 @@ func (m minerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, i
 
 	wrappedStore := adt11.WrapStore(ctx, store)
 
-	newSectors, err := migrateSectorsWithCache(ctx, wrappedStore, in.Cache, in.Address, inState.Sectors)
+	newSectors, err := m.migrateSectorsWithCache(ctx, wrappedStore, in.Cache, in.Address, inState.Sectors)
```
arajasek, line 152: (not related to this PR) 

the portion above this should be dropped entirely -- we are not migrating post proof types / minerinfos in this network upgrade.

arajasek (commented):
Bunch of minor comments, mostly

builtin/v12/migration/miner.go
```diff
@@ -171,11 +213,18 @@ func migrateSectorsWithCache(ctx context.Context, store adt11.Store, cache migra
 		if err != nil {
 			return cid.Undef, xerrors.Errorf("failed to get previous outRoot from cache: %w", err)
 		}
+
+		okSectorIndexOut, prevSectorIndexRoot, err := cache.Read(migration.MinerPrevSectorDealIndexKey(inRoot))
+		if err != nil {
+			return cid.Undef, xerrors.Errorf("failed to get previous outRoot from cache: %w", err)
```
arajasek, line 219: copy pasta error log

builtin/v12/migration/miner.go
```diff
@@ -171,11 +213,18 @@ func migrateSectorsWithCache(ctx context.Context, store adt11.Store, cache migra
 		if err != nil {
 			return cid.Undef, xerrors.Errorf("failed to get previous outRoot from cache: %w", err)
 		}
+
+		okSectorIndexOut, prevSectorIndexRoot, err := cache.Read(migration.MinerPrevSectorDealIndexKey(inRoot))
```
arajasek, line 217: I think you want to look up `prevInRoot` here, not `inRoot`. We won't have a cached result for `inRoot` but we _should_ have one for `prevInRoot`.

arajasek: (By looking up `inRoot` I think you'll wind up doing the migration from scratch every time)

snissn: changed here #a008d024f687a6fc24fc0ebac27b4961037343a5

builtin/v12/migration/top.go
```diff
@@ -95,25 +112,91 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
 	if err != nil {
 		return cid.Undef, xerrors.Errorf("failed to run migration: %w", err)
 	}
 
-	outCid, err := actorsOut.Flush()
+	// Retrieve the sector deal IDs from the minerMigrator. These IDs are crucial for the migration of market actor state.
+	sectorDealIDs, err := minerMigrator.sectorDeals.Root()
 	if err != nil {
-		return cid.Undef, xerrors.Errorf("failed to flush actorsOut: %w", err)
+		return cid.Undef, err
```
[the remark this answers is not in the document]
snissn: #190fe18fab4b1ce540c0a89b4ef13325124dc8b6  thx changed

builtin/v12/migration/top.go
```diff
@@ -95,25 +112,91 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
+
+	//save sectorDealIdIndex in Cache
+	if err = cache.Write(migration.MarketSectorIndexKey(), sectorDealIDs); err != nil {
+		return cid.Undef, xerrors.Errorf("failed to write inkey to cache: %w", err)
+	}
+
+	// Create the new state
+	newMarketState := market12.State{
+		Proposals:                     oldMarketState.Proposals,
+		States:                        oldMarketState.States, // FIXME migrate deal states to include sector number
```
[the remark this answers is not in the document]
snissn: from git blame - i think this is related to alex's work, but we have one of his commits in this branch for the release workflow

606d9d5c (Alex North     2023-09-06 08:47:57 +1000 171)                 States:
                       oldMarketState.States, // FIXME migrate deal states to in
clude sector number

builtin/v12/migration/top.go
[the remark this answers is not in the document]
snissn: #190fe18fab4b1ce540c0a89b4ef13325124dc8b6

builtin/v12/migration/miner.go
```diff
@@ -224,65 +272,111 @@ delta = prevInRoot - inRoot
 
 merge( prevOutRoot, Migrated(delta) )
 */
-func migrateSectorsWithDiff(ctx context.Context, store adt11.Store, inRoot cid.Cid, prevInRoot cid.Cid, prevOutRoot cid.Cid) (cid.Cid, error) {
-	// we have previous work, but the AMT has changed -- diff them
+
+// todo modify hamt based on diff
```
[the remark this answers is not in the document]
snissn: done :D

builtin/v12/migration/miner.go
snissn: a9745f4 fixed

builtin/v12/migration/miner.go
snissn: 4f851d59c1b66e9a2fb40cede2eb4d648a3e38e3

builtin/v12/migration/miner.go
```diff
@@ -224,65 +272,111 @@ delta = prevInRoot - inRoot
...
+func (m minerMigrator) migrateSectorsWithDiff(ctx context.Context, store adt11.Store, minerAddr address.Address, inRoot cid.Cid, prevInRoot cid.Cid, prevOutRoot cid.Cid, prevSectorIndexRoot cid.Cid) (cid.Cid, cid.Cid, error) {
```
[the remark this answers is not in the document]
snissn: sectorToDealIDHamt?

builtin/v12/migration/miner.go
snissn: thanks, changed! This was pointed out by zen also 0a40559

builtin/v12/migration/top.go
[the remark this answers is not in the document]
arajasek: Yes, but as far as I understand, this work is still outstanding. We need to do the second portion of the migration described [here](https://github.com/filecoin-project/FIPs/blob/8aee21c54d213dbba71102b91d4ae0d31df0e2f9/FIPS/fip-0076.md#backwards-compatibility) (`and to add a sector number to and remove allocation ID from each DealState`).

builtin/v12/migration/miner.go
arajasek: @snissn Can you make this change too (ideally as a pre-factor that we land first, but otherwise just as part of this PR)?

arajasek: I'm worried about forgetting about it.

builtin/v12/migration/miner.go
```diff
@@ -67,12 +74,34 @@ func newMinerMigrator(ctx context.Context, store cbor.IpldStore, outCode cid.Cid
 
 	}
 
+	//load sector index from datastore, or create a new one if not found
+	okSectorIndex, prevSectorIndexRoot, err := cache.Read(migration.MarketSectorIndexKey())
```
arajasek, line 78: We could consider reading this in `top.go`, and just passing `prevSectorIndexRoot` here so that we don't pass the entire cache here. I don't think it matters too much.

builtin/v12/migration/miner.go
```diff
@@ -67,12 +74,34 @@ func newMinerMigrator(ctx context.Context, store cbor.IpldStore, outCode cid.Cid
...
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get previous sector index from cache: %w", err)
+	}
+	var sectorDeals *adt12.Map
+	if okSectorIndex {
+		sectorDeals, err = adt12.AsMap(ctxStore, prevSectorIndexRoot, builtin.DefaultHamtBitwidth)
+		if err != nil {
+			return nil, xerrors.Errorf("reading cached sectorDeals tree: %w", err)
+		}
+		fmt.Println("Loaded HAMT from cache: ", prevSectorIndexRoot)
```
arajasek, line 88: drop?

builtin/v12/migration/miner.go
```diff
@@ -67,12 +74,34 @@ func newMinerMigrator(ctx context.Context, store cbor.IpldStore, outCode cid.Cid
...
+	} else {
+		// New mapping of sector IDs to deal IDS, grouped by storage provider.
+		sectorDeals, err = adt12.MakeEmptyMap(ctxStore, builtin.DefaultHamtBitwidth)
+		if err != nil {
+			return nil, xerrors.Errorf("creating new state tree: %w", err)
```
arajasek, line 93: bad error message

builtin/v12/migration/miner.go
```diff
@@ -120,11 +149,24 @@ func (m minerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, i
...
+	if err != nil {
+		return nil, xerrors.Errorf("failed to read sector to deal id hamt from cache: %w", err)
+	}
+	if !okCacheRead {
+		return nil, xerrors.Errorf("sector to deal id hamt not found in cache for address: %s", in.Address)
+	}
+	//add to new address sector deal id hamt index
+	err = m.addSectorToDealIDHamtToSectorDeals(sectorToDealIdHamtCid, in.Address)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to add sector to deal id hamt to for minerAddr to HAMT: %w", err)
```
arajasek, line 167: bad error msg

arajasek (commented):
More notes.

builtin/v12/migration/miner.go
```diff
@@ -224,65 +280,112 @@ delta = prevInRoot - inRoot
...
+
+	sectorToDealIdHamt, err := adt12.MakeEmptyMap(store, builtin.DefaultHamtBitwidth)
+	if err != nil {
+		return nil, cid.Undef, xerrors.Errorf("creating new state tree: %w", err)
 	}
 
 	var sectorInfo miner11.SectorOnChainInfo
 	if err = inArray.ForEach(&sectorInfo, func(k int64) error {
+
+		err = addSectorNumberToDealIdHAMT(sectorToDealIdHamt, sectorInfo, store)
```
arajasek, line 375: check error here

arajasek (commented):
(forgot to submit review yday)

arajasek:
Deferred to nv22.

PR #212, opened 2023-09-01
replace SimpleQaPower with flags

LesnyRumcajs:
See https://github.com/filecoin-project/builtin-actors/issues/1275
and https://github.com/filecoin-project/builtin-actors/pull/1395 (most notably @ZenGround0 [comment](https://github.com/filecoin-project/builtin-actors/pull/1395#pullrequestreview-1605376348))

In short, we want to have an easily extensible flags field for the SectorOnChainInfo instead of a field per flag, which is wasteful.

### DIFF
--- a/builtin/v12/migration/miner.go
+++ b/builtin/v12/migration/miner.go
@@ -372,16 +372,16 @@ func (m minerMigrator) migrateDeadlines(ctx context.Context, store adt11.Store, 
 }
 
 func migrateSectorInfo(sectorInfo miner11.SectorOnChainInfo) *miner12.SectorOnChainInfo {
-  // For a sector that has not been updated: the Activation is correct and ReplacedSectorAge is zero.
-  // For a sector that has been updated through SnapDeals: Activation is the epoch at which it was upgraded, and ReplacedSectorAge is delta since the true activation.
-  // For a sector that has been updated through the old CC path: Activation is correct
-  // Thus, we want to set:
-  //
-  // PowerBaseEpoch := Activation (in all cases)
-  // Activation := Activation (for non-upgraded sectors and sectors upgraded through old CC path)
-  // Activation := OldActivation - ReplacedSectorAge (for sectors updated through SnapDeals)
-  //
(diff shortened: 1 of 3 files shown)


builtin/v12/migration/miner.go
```diff
@@ -372,21 +372,28 @@ func (m minerMigrator) migrateDeadlines(ctx context.Context, store adt11.Store,
...
 func migrateSectorInfo(sectorInfo miner11.SectorOnChainInfo) *miner12.SectorOnChainInfo {
-	// For a sector that has not been updated: the Activation is correct and ReplacedSectorAge is zero.
-	// For a sector that has been updated through SnapDeals: Activation is the epoch at which it was upgraded, and ReplacedSectorAge is delta since the true activation.
-	// For a sector that has been updated through the old CC path: Activation is correct
-	// Thus, we want to set:
-	//
-	// PowerBaseEpoch := Activation (in all cases)
-	// Activation := Activation (for non-upgraded sectors and sectors upgraded through old CC path)
-	// Activation := OldActivation - ReplacedSectorAge (for sectors updated through SnapDeals)
+  // For a sector that has not been updated: the Activation is correct and ReplacedSectorAge is zero.
```
ZenGround0, line 375: nit: Indentation is off here and elsewhere

builtin/v12/miner/miner_state.go
```diff
@@ -158,23 +158,30 @@ type SectorPreCommitOnChainInfo struct {
 	PreCommitEpoch   abi.ChainEpoch
 }
 
+type SectorOnChainInfoFlags int64
```
ZenGround0, line 161: nit: probably slightly better as a uint64

ZenGround0 (approved):
Two small things but otherwise LGTM

LesnyRumcajs:
Hm, should the type of the flag be aligned with the type in bultin actors?

ZenGround0:
> Hm, should the type of the flag be aligned with the type in bultin actors? I.e., both either 32 or 64 bits?

I gave it a try and cbor gen does not work with int32s.  This should be fine for the long foreseeable future until with have 31 new flags.  I'll file an issue.

PR #213, opened 2023-09-02
Sector info flags: workaround permissions

ZenGround0:
This PR is a duplicate of #212 which is already approved.  We are in an annoying situation where CI jobs from forks were disabled and the author of #212 pushed from a fork and is now afk for a long time.  I enabled building from forks but it doesn't retroactively apply without a push so the next best thing is to file a duplicate and ask for an approval.


ZenGround0:
aha just pushing this was enough to get CI on the original to pass, all set

PR #215, opened 2023-09-05
Update deal state schema to match Rust

anorth:
#210 doesn't yet include updating the market DealState schema. This PR isn't a complete resolution (I'm no longer familiar with the rest of the migration code), but intended to be cherry-picked or otherwise cloned into #210 as a prompt to migrate those objects.

See https://github.com/filecoin-project/builtin-actors/pull/1347 and https://github.com/filecoin-project/builtin-actors/pull/1403


anorth:
Note that I have not ported changes to the state checks from Rust to Go. My understanding is that these have drifted in many places because no-one is maintaining the Go version. This is a predictable result of the code duplication involved. However, it would be good to prevent the problem getting worse.

snissn (commented):
looks good

PR #216, opened 2023-09-06
Fix ProveCommit2Return and DealState structs in GST

shrenujbansal:
- ProveCommit2Return should have a single BatchReturn field (not an array) based on actors code
- DealState adds the SectorNumber field and removes the VerifiedClaim field


snissn (approved):
looks good, 
confirming that 
    ProveCommit2Return returns a single BatchReturn field (not an array) 
and 
    DealState adds the SectorNumber field and removes the VerifiedClaim field

PR #220, opened 2023-09-23
fix: v12: ComputeDataCommitment is deprecated

arajasek:
Removed in https://github.com/filecoin-project/builtin-actors/pull/1354


builtin/v12/market/methods.go
```diff
@@ -17,7 +17,7 @@ var Methods = map[abi.MethodNum]builtin.MethodMeta{
 	5: {"VerifyDealsForActivation", *new(func(*VerifyDealsForActivationParams) *VerifyDealsForActivationReturn)}, // VerifyDealsForActivation
 	6: {"ActivateDeals", *new(func(*ActivateDealsParams) *abi.EmptyValue)},                                       // ActivateDeals
 	7: {"OnMinerSectorsTerminate", *new(func(*OnMinerSectorsTerminateParams) *abi.EmptyValue)},                   // OnMinerSectorsTerminate
-	8: {"ComputeDataCommitment", *new(func(*ComputeDataCommitmentParams) *ComputeDataCommitmentReturn)},          // ComputeDataCommitment
+	8: {"ComputeDataCommitment", nil},                                                                            // deprecated
```
magik6k, line 20: Do we also want to change this string to DeprecatedSomethingSomething? I'm fine with either way.

magik6k (approved):

builtin/v12/market/methods.go
Stebalien: It looks like that's not what we currently do, so let's stick with the current convention for now.

PR #221, opened 2023-09-23
fix: use the correct type for v12 VerifyDealsForActivationReturn

Stebalien:
Unfortunately, this also means we need to update cbor-gen with a large refactor.


arajasek (approved):

arajasek:
Needed for the change in https://github.com/filecoin-project/builtin-actors/pull/1361

PR #222, opened 2023-09-26
fix: v12: Update DealMaxDuration to 1278 days

rjan90:
Update DealMaxDuration to 1278 days to be in-line with [FIP-0052](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0052.md).


arajasek (approved):
Thank you!

PR #230, opened 2023-11-08
fix: invariants: use the power base epoch instead of the activation

Stebalien:
Otherwise, we compute the wrong expected power and expected deal start times.


builtin/v12/miner/invariants.go
```diff
@@ -95,7 +95,7 @@ func CheckStateInvariants(st *State, store adt.Store, balance abi.TokenAmount) (
 
 			for _, dealID := range sector.DealIDs {
 				minerSummary.Deals[dealID] = DealSummary{
-					SectorStart:      sector.Activation,
+					SectorStart:      sector.PowerBaseEpoch,
```
Stebalien, line 98: I'm less sure about this, but I _think_ this is what we want.

Stebalien:
I also added a commit to _not- assert quantization because we no longer guarantee that.

builtin/v12/miner/invariants.go
Stebalien: It's not.

anorth (approved):

PR #232, opened 2023-11-20
[WIP] feat: Add nv22 skeleton

TippyFlitsUK:
Addition of Network Version 22 skeleton


aarshkshah1992:
@TippyFlitsUK What script did you use to generate the changes in this PR ?

TippyFlitsUK:
I was following the guide [here](https://github.com/filecoin-project/lotus/blob/9cf748215c001d641a3584e070bfb6d4b1ef8b1b/documentation/misc/Building_a_network_skeleton.md) @aarshkshah1992 👍 
I believe it was the `make gen` step.

network/version.go
```diff
@@ -29,6 +29,7 @@ const (
 	Version19                 // Lightning
 	Version20                 // Thunder
 	Version21                 // SyntheticPoRep
+	Version22                 // DDO
```
arajasek, line 32: Nit: Call this TBD, and call the line above Watermelon?

builtin/v13/migration/miner.go
```diff
@@ -0,0 +1,364 @@
+package migration
```
arajasek, line 1: This file is specific to the nv21 migration. I would propose dropping the entire `migration` directory for now -- we'll have to add the "lightweight" migration, but we can land this PR without that.

arajasek (commented):
Just a couple nits.

arajasek (approved):

PR #234, opened 2023-12-05
feat: add simple nv22 migration

arajasek:
"Lightest" migration that updates Code IDs based on new manifest, and the state of the system actor accordingly.


magik6k (approved):

arajasek:
(merging with broken CI since CI is broken on master anyway, will get fixed in subsequent PR).

PR #209, opened 2023-08-28
wip nv12 direct data onboarding types

magik6k:
This PR adds Go types compatible with direct-data-onboarding types defined in https://github.com/filecoin-project/builtin-actors/pull/1380/files

PR #236, opened 2023-12-08
DDO types, nv22 edition

magik6k:
For https://github.com/filecoin-project/lotus/pull/11226
Resurrects https://github.com/filecoin-project/go-state-types/pull/209 for nv22
On top of https://github.com/filecoin-project/go-state-types/pull/234


builtin/v13/migration/miner.go
```diff
@@ -0,0 +1,105 @@
...
+	"golang.org/x/xerrors"
+	"sync"
+)
+
+type providerSectors struct {
+	lk sync.Mutex
+
+	//providerSectors map[address.Address]cid.Cid // HAMT[SectorNumber]SectorDealIDs
+	dealToSector       map[abi.DealID]abi.SectorID
+	minerToSectorDeals map[address.Address]map[abi.SectorNumber][]abi.DealID
```
magik6k, line 22: This map is unused and should be removed

builtin/v13/miner/methods.go
```diff
@@ -59,4 +59,7 @@ var Methods = map[abi.MethodNum]builtin.MethodMeta{
 	builtin.MustGenerateFRCMethodNum("GetVestingFunds"):      {"GetVestingFundsExported", *new(func(*abi.EmptyValue) *GetVestingFundsReturn)},                              // GetVestingFundsExported
 	builtin.MustGenerateFRCMethodNum("GetPeerID"):            {"GetPeerIDExported", *new(func(*abi.EmptyValue) *GetPeerIDReturn)},                                          // GetPeerIDExported
 	builtin.MustGenerateFRCMethodNum("GetMultiaddrs"):        {"GetMultiaddrsExported", *new(func(*abi.EmptyValue) *GetMultiAddrsReturn)},                                  // GetMultiaddrsExported
+	// 33 MovePartitions
+	34: {"ProveCommitSectors3", *new(func(*ProveCommitSectors3Params) *ProveCommitSectors3Return)},    // ProveCommitSectors3
+	35: {"ProveReplicaUpdates3", *new(func(*ProveReplicaUpdates3Params) *ProveReplicaUpdates3Return)}, // ProveReplicaUpdates3
```
Stebalien, line 64: I would kind of prefer to put private methods next to each other. But I guess that's a future update.

builtin/v13/miner/miner_types.go
```diff
@@ -394,3 +394,117 @@ type GetPeerIDReturn struct {
...
+
+type BatchReturn struct {
+	SuccessCount uint64
+	FailCodes    []FailCode
+}
+
+type FailCode struct {
+	// Idx represents the index of the operation that failed within the batch.
+	Idx  uint64
+	Code xc.ExitCode // todo correct?
```
Stebalien, line 448: Yes.

builtin/v13/miner/miner_types.go
```diff
@@ -394,3 +394,117 @@ type GetPeerIDReturn struct {
...
+type PieceActivationReturn struct {
+	Claimed       bool
+	Notifications []DataActivationNotificationReturn
+}
+
+// DataActivationNotificationReturn represents the return value for a data activation notification.
+type DataActivationNotificationReturn struct {
+	Code xc.ExitCode // todo correct??
+	Data []byte
+}
```
Stebalien, line 461: I think these were dropped at some point? Unless I missed something.

Stebalien (changes requested):

builtin/v13/miner/miner_types.go
magik6k: Yeah, afaict those got replaced with SectorChanges

Stebalien (changes requested):
The state-types look good, but I'm concerned about the migration logic.

builtin/v13/migration/top.go
```diff
@@ -67,7 +66,16 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
 	// Set of prior version code CIDs for actors to defer during iteration, for explicit migration afterwards.
 	deferredCodeIDs := make(map[cid.Cid]struct{})
```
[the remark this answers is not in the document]
magik6k: Removed in https://github.com/filecoin-project/go-state-types/pull/240

builtin/v13/market/market_state.go
```diff
@@ -55,6 +56,14 @@ type State struct {
 
 	// Verified registry allocation IDs for deals that are not yet activated.
 	PendingDealAllocationIds cid.Cid // HAMT[DealID]AllocationID
+
+	/// Maps providers to their sector IDs to deal IDs.
+	/// This supports finding affected deals when a sector is terminated early
+	/// or has data replaced.
+	/// Grouping by provider limits the cost of operations in the expected use case
+	/// of multiple sectors all belonging to the same provider.
+	/// HAMT[ActorID]HAMT[SectorNumber]SectorDealIDs
```
rvagg, line 65:
```suggestion
	/// HAMT[ActorID]HAMT[SectorNumber][]DealID
```

the FIP now has it described in this form

rvagg: actually I guess the named type makes more sense here so maybe it's reasonable to leave it like this

builtin/v13/miner/miner_types.go
```diff
@@ -394,3 +394,105 @@ type GetPeerIDReturn struct {
 type GetMultiAddrsReturn struct {
 	MultiAddrs []byte
 }
+
+// ProveCommitSectors3Params represents the parameters for proving committed sectors.
+type ProveCommitSectors3Params struct {
+	SectorActivations          []SectorActivationManifest
+	SectorProofs               [][]byte
+	AggregateProof             []byte
+	AggregateProofType         abi.RegisteredAggregationProof
```
rvagg, line 403:
```suggestion
	AggregateProofType         *abi.RegisteredAggregationProof
```

probably, with FIP update, https://github.com/filecoin-project/builtin-actors/pull/1511

builtin/v13/miner/miner_types.go
```diff
@@ -394,3 +394,105 @@ type GetPeerIDReturn struct {
...
+	Code xc.ExitCode // todo correct?
+}
+
+// ProveReplicaUpdates3Params represents the parameters for proving replica updates.
+type ProveReplicaUpdates3Params struct {
+	SectorUpdates              []SectorUpdateManifest
+	SectorProofs               [][]byte
+	AggregateProof             []byte
+	UpdateProofsType           abi.RegisteredUpdateProof
+	AggregateProofType         abi.RegisteredAggregationProof
```
rvagg, line 457:
```suggestion
	AggregateProofType         *abi.RegisteredAggregationProof
```

probably https://github.com/filecoin-project/builtin-actors/pull/1511

builtin/v13/miner/miner_types.go
magik6k: Will be a separate update when that PR lands

migration/util.go
```diff
@@ -204,3 +205,45 @@ func CachedMigration(cache MigrationCache, m ActorMigration) ActorMigration {
...
+	d.ds.lk.Lock()
+	defer d.ds.lk.Unlock()
+
+	_, ok := d.ds.todo[in.Address]
+	if ok {
+		return nil, xerrors.Errorf("actor %s already in deferred set", in.Address)
+	}
+	d.ds.todo[in.Address] = struct{}{}
+
+	return nil, errMigrationDeferred
```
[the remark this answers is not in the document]
magik6k: In #240 to avoid rebasing - https://github.com/filecoin-project/go-state-types/pull/240/commits/7e39f88cd964d598a558f382075dd48a9c401003

(maybe makes sense to just merge 240 here now, but separate is probably easier to review)

arajasek (changes requested):
Please add new v13 invariants that cover the new ProviderSectors field. You can see the `CheckDealStatesAgainstSectors` ([here](https://github.com/filecoin-project/go-state-types/blob/f4366fdb96de135e00911166c1b461f0c2bdd52e/builtin/v13/check.go#L254) method for what this looks like. It shouldn't be too hard, though you might have to make some additions to the miner's `StateSummary` / `DealSummary`. We can't really have full confidence in the migration without that.

Otherwise, this generally looks good to me.

migration/util.go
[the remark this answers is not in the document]
Stebalien: Can we just move the entire migration to that PR? If we can't make changes to this PR, the review process is kind of broken. That and the other PR makes _so_ many changes to the migration that I'm not sure if there's even a point of reviewing it here.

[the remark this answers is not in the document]
magik6k: Everything is here now

builtin/v13/market/cbor_gen.go
```diff
@@ -441,23 +472,84 @@ func (t *DealState) UnmarshalCBOR(r io.Reader) (err error) {
...
+}
 
-	{
+func (t *SectorDealIDs) MarshalCBOR(w io.Writer) error {
+	cw := cbg.NewCborWriter(w)
 
-		maj, extra, err = cr.ReadHeader()
-		if err != nil {
+	// (*t) (market.SectorDealIDs) (slice)
+	if len((*t)) > cbg.MaxLength {
```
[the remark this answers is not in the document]
magik6k: Ah yes. Imo not worth blocking on that as it doesn't really matter for the types we have here, but I'll put this as a todo somewhere near the top of my todo list

builtin/v13/migration/market.go
```diff
@@ -0,0 +1,503 @@
...
+	var newState market13.DealState
+	err = oldStateArray.ForEach(&oldState, func(i int64) error {
+		deal := abi.DealID(i)
+
+		newState.SlashEpoch = oldState.SlashEpoch
+		newState.LastUpdatedEpoch = oldState.LastUpdatedEpoch
+		newState.SectorStartEpoch = oldState.SectorStartEpoch
+		newState.SectorNumber = 0
+
+		if oldState.SectorStartEpoch != -1 {
```
arajasek, line 437: Can we do exactly what the FIP says to do:

- check whether `LastUpdatedEpoch` is -1
- if yes, find the sector in `providerSectors` (if you can't assert that the deal is past EndEpoch or fail)
- if no, just use 0

builtin/v13/check.go
```diff
@@ -288,6 +288,67 @@ func CheckDealStatesAgainstSectors(acc *builtin.MessageAccumulator, minerSummari
 		acc.Require(deal.SlashEpoch <= sectorDeal.SectorExpiration,
 			"deal state slashed at %d after sector expiration %d for miner %v",
 			deal.SlashEpoch, sectorDeal.SectorExpiration, deal.Provider)
+
+		acc.Require(deal.SectorNumber == sectorDeal.SectorNumber,
```
arajasek, line 292: You probably need to exclude `deal.SectorNumber == 0` here (ideally it would be `(deal.SectorNumber == 0 && deal.LastUpdatedEpoch > currEpoch)`).

builtin/v13/check.go
```diff
@@ -288,6 +288,67 @@ func CheckDealStatesAgainstSectors(acc *builtin.MessageAccumulator, minerSummari
...
+		}
+		for dealID, dealSummary := range summary.Deals {
+			_, found := inverseSectorDeals[dealID]
+			acc.Require(!found, "deal %d found in multiple sectors", dealID)
+
+			inverseSectorDeals[dealID] = abi.SectorID{Miner: abi.ActorID(mid), Number: dealSummary.SectorNumber}
+		}
+	}
+
+	for dealID, sectorID := range inverseSectorDeals {
```
arajasek, line 342: The finer points of this are going over my head -- we're going to need @anorth's help stating exactly what is possible and not possible here (or equivalent invariants in Rust that we can pull in here).

arajasek (commented):
Some notes

builtin/v13/migration/market.go
magik6k: > LastUpdatedEpoch is -1

`is` or `is not`? My belief is that LastUpdatedEpoch is only -1 before the deal is activated for the first time, so that sounds wrong.

builtin/v13/check.go
magik6k: Added the condition

builtin/v13/check.go
magik6k: I've added this because what it can catch seemed like an incorrect behavior, but yes, need more input on that.

builtin/v13/migration/market.go
```diff
@@ -0,0 +1,536 @@
...
+			newState.LastUpdatedEpoch = oldState.LastUpdatedEpoch
+			newState.SectorStartEpoch = oldState.SectorStartEpoch
+			newState.SectorNumber = 0 // terminated / not found (?)
+
+			if oldState.SlashEpoch == -1 {
+				si, err := addProviderSectorEntry(deal, &newState)
+				if err != nil {
+					return cid.Cid{}, cid.Cid{}, xerrors.Errorf("failed to add provider sector entry: %w", err)
+				}
+				newState.SectorNumber = si
```
LesnyRumcajs, line 212: My Go-Fu may be lacking but doesn't it make the `newState.SectorNumber = sid.Number` in `addProviderSectorEntry` redundant?

go.mod
```diff
@@ -20,8 +20,26 @@ require (
 	github.com/multiformats/go-multihash v0.2.1
 	github.com/multiformats/go-varint v0.0.6
 	github.com/stretchr/testify v1.7.0
-	github.com/whyrusleeping/cbor-gen v0.0.0-20230923211252-36a87e1ba72f
+	github.com/whyrusleeping/cbor-gen v0.0.0-20240124134906-0efd71a8493b
```
anorth, line 23: Maybe this could have been a separate PR to remove all the noise from this one?

builtin/v13/check.go
```diff
@@ -288,6 +288,67 @@ func CheckDealStatesAgainstSectors(acc *builtin.MessageAccumulator, minerSummari
...
 			"deal state slashed at %d after sector expiration %d for miner %v",
 			deal.SlashEpoch, sectorDeal.SectorExpiration, deal.Provider)
+
+		acc.Require((deal.SectorNumber == sectorDeal.SectorNumber) || (deal.SectorNumber == 0 && deal.LastUpdatedEpoch != -1),
+			"deal sector number %d does not match sector %d for miner %v (ds: %#v; ss %#v)",
+			deal.SectorNumber, sectorDeal.SectorNumber, deal.Provider, deal, sectorDeal)
+	}
+
+	//	/// HAMT[ActorID]HAMT[SectorNumber]SectorDealIDs
+	foundSectorDeals := make(map[abi.DealID]abi.SectorID)
```
anorth, line 298:
```suggestion
	marketDealToSector := make(map[abi.DealID]abi.SectorID)
```

builtin/v13/check.go
```diff
@@ -288,6 +288,67 @@ func CheckDealStatesAgainstSectors(acc *builtin.MessageAccumulator, minerSummari
...
+			acc.Require(found, "deal %d not found in miner %v for sector %v", dealID, maddr, sectorID)
+
+			_, found = foundSectorDeals[dealID]
+			acc.Require(!found, "deal %d found in multiple sectors", dealID)
+
+			foundSectorDeals[dealID] = sectorID
+		}
+	}
+
+	inverseSectorDeals := make(map[abi.DealID]abi.SectorID)
```
anorth, line 327:
```suggestion
	minerDealToSector := make(map[abi.DealID]abi.SectorID)
```
i.e. it's from the miner state

builtin/v13/check.go
```diff
@@ -288,6 +288,67 @@ func CheckDealStatesAgainstSectors(acc *builtin.MessageAccumulator, minerSummari
...
+	for dealID, sectorID := range inverseSectorDeals {
+		ds := marketSummary.Deals[dealID]
+		_, found := marketSummary.ProviderSectors[sectorID]
+		if ds == nil {
+			acc.Require(!found, "expired deal %d found in sector %v", dealID, sectorID) // TODO we see this, is this fine?
+
+			continue // expired deal
+		}
+
+		acc.Require(found, "deal %d found in sector %v not found in market (ds: %v)", dealID, sectorID, ds) // TODO we see this, is this fine?
```
anorth, line 351: Is the intent here to check that a deal ID found in SectorOnChainInfo has a corresponding entry for the (miner, sector) in the market ProviderSectors? This is valid only for non-terminated deals. You need to check the deal state first.

anorth (commented):
I have only reviewed the types and invariant check so far, not the migration code.

The invariant checks would be a lot easier to review if there were comments describing the intent of the check. Then I could check that the intent is correct, and that the implementation matches that intent, separately.

FIP-0076 is decoupling miner and market state, so in general there won't be many invariants between them. There are two different things you want to check here:
- Test that the migration is correct. This can involved some checks that market actor state (deal sector numbers) implies some miner actor state, since the former was supposed to be derived in part from the latter.
- Test that state invariants are maintained in the future. SectorOnChainInfo deal ids are deprecated, so there can be few (no?) useful invariants checked there. Checking that sectors are/not terminated is probably the only cross-actor thing you can do, but there is probably more checking internal to the market actor.

builtin/v13/check.go
```diff
@@ -288,6 +288,67 @@ func CheckDealStatesAgainstSectors(acc *builtin.MessageAccumulator, minerSummari
...
+			acc.Addf("provider %v for sector %v not found among miners", sectorID, sectorID)
+			continue
+		}
+
+		sector := minerSummary.SectorsWithDeals[sectorID.Number]
+		acc.Require(sector, "sector %v not found in miner %v for deals %v", sectorID, maddr, dealIDs)
+
+		for _, dealID := range dealIDs {
+			_, found := minerSummary.Deals[dealID]
+			acc.Require(found, "deal %d not found in miner %v for sector %v", dealID, maddr, sectorID)
```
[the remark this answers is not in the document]
magik6k: Correct, this is just a migration correctness assert. I'll prefix the message with MIGRATION-ONLY to make it clearer that this assert is expected to break post-migration

builtin/v13/market/invariants.go
```diff
@@ -258,6 +257,47 @@ func CheckStateInvariants(st *State, store adt.Store, balance abi.TokenAmount, c
...
+				copy(dealIDsCopy, dealIDs)
+
+				providerSectors[abi.SectorID{Miner: abi.ActorID(provider), Number: abi.SectorNumber(sectorNumber)}] = dealIDsCopy
+
+				// check against proposalStats
+				for _, dealID := range dealIDsCopy {
+					st, found := proposalStats[dealID]
+					acc.Require(found, "deal id %d in provider sectors not found in proposals", dealID)
+					acc.Require(st.SectorNumber == abi.SectorNumber(sectorNumber), "deal id %d sector number %d does not match sector id %d", dealID, st.SectorNumber, sectorNumber)
+				}
```
anorth, line 291: This check should be made in the opposite direction too, perhaps even more importantly. Every deal state that is not activated and not terminated/slashed should find that sector number in the providerSectors map.

Thought: it might be easy to check this by collecting the tuples from each collection while building them, then asserting equality of the sets.

builtin/v13/check.go
```diff
@@ -288,6 +288,61 @@ func CheckDealStatesAgainstSectors(acc *builtin.MessageAccumulator, minerSummari
...
+
+			provID, err := address.IDFromAddress(deal.Provider)
+			if err != nil {
+				acc.Addf("error creating ID address: %v", err)
+				continue
+			}
+
+			acc.Require(abi.ActorID(provID) == sectorID.Miner, "deal %d has provider %v, expected %v", dealID, deal.Provider, sectorID)
+			acc.Require(deal.SectorNumber == sectorID.Number, "deal %d has sector number %d, expected %d", dealID, deal.SectorNumber, sectorID.Number)
+		}
```
anorth, line 345: This check is internal to the market actor now, so should move into market/invariants.go.

builtin/v13/migration/market.go
```diff
@@ -0,0 +1,590 @@
...
+	}
+
+	// in-memory maps with changesets to be applied to prevOutProviderSectors
+	providerSectorsMem := map[abi.ActorID]map[abi.SectorNumber][]abi.DealID{}        // added
+	providerSectorsMemRemoved := map[abi.ActorID]map[abi.SectorNumber][]abi.DealID{} // removed
+
+	addProviderSectorEntry := func(deal abi.DealID, newState *market13.DealState) (abi.SectorNumber, error) {
+		sid, ok := m.providerSectors.dealToSector[deal]
+		if !ok {
+			return 0, xerrors.Errorf("deal %d not found in providerSectors", deal) // todo is this normal and possible??
```
arajasek, line 164: I don't think so, but I'm not sure.

arajasek (approved):
Some notes.

builtin/v13/check.go
magik6k: Moved into market checks

builtin/v13/migration/miner.go
```diff
@@ -0,0 +1,229 @@
...
+							return nil, err
+						}
+						pjson, err := json.MarshalIndent(sectorAfter, "", "  ")
+						if err != nil {
+							return nil, err
+						}
+
+						fmt.Println("sector before: ", string(pjsonb))
+						fmt.Println("sector after: ", string(pjson))
+						return nil, xerrors.Errorf("WHAT?! sector %d modified, this not supported and not supposed to happen", i) // todo: is it? Can't happen w/o a deep, deep reorg, and even then we wouldn't use the cache??
```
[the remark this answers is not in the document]
magik6k: Do we even use the cache through reorgs? If not this really can't happen, and if we do use the cache through reorgs I imagine the market migrator makes a lot of invalid assumptions too.

(also this reminds me that removedDealToSector is actually no longer used)

builtin/v13/migration/miner.go
```diff
@@ -0,0 +1,229 @@
...
+							Miner:  abi.ActorID(mid),
+							Number: sectorNo,
+						}
+					}
+					m.providerSectors.lk.Unlock()
+				}
+
+				// extensions, etc. here; we don't care about those
+			case amt.Remove:
+				// related deals will also get removed in the market, so we don't have anything to do here
```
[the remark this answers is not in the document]
magik6k: Clarified

builtin/v13/migration/market.go
```diff
@@ -0,0 +1,590 @@
...
+			if oldState.SlashEpoch != -1 && prevOldState.SlashEpoch == -1 {
+				// not slashed -> slashed
+				//fmt.Printf("deal %d slash -1 -> %d\n", deal, oldState.SlashEpoch)
+
+				if err := removeProviderSectorEntry(deal, &newState); err != nil {
+					return cid.Cid{}, cid.Cid{}, xerrors.Errorf("failed to remove provider sector entry: %w", err)
+				}
+			}
+
+			/*if oldState.SectorStartEpoch == -1 && prevOldState.SectorStartEpoch != -1 {
```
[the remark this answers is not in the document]
magik6k: Keeping the commented prints for now, if we'll need to debug this just uncommenting should give us all the information neeed, but will remove this block of code.

builtin/v13/migration/miner.go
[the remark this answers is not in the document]
arajasek: Yes, we use the cache through reorgs, and should be fully reorg-resistant. There is no assumption that the state it ran on is an ancestor of the current state tree. Instead, the assumption is just that if we cached that X -> Y, we can use that information to calculate X' -> Y', where X and X' probably have some commonalities.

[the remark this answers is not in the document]
arajasek: What are some of the invalid assumptions you're concerned about? 

We _could_ tweak how Lotus uses the cache to wipe it if we know its base state was re-orged out, but I'd rather not do that.

builtin/v13/migration/market.go
```diff
@@ -0,0 +1,548 @@
...
+
+				// FIP: add the deal ID to the ProviderSectors mapping for the provider's actor ID and sector number.
+				if _, ok := providerSectorsMem[sid.Miner]; !ok {
+					providerSectorsMem[sid.Miner] = make(map[abi.SectorNumber][]abi.DealID)
+				}
+				providerSectorsMem[sid.Miner][sid.Number] = append(providerSectorsMem[sid.Miner][sid.Number], deal)
+			} else {
+				//fmt.Printf("deal %d not found in providerSectors: %v\n", deal, oldState)
+
+				newState.SectorNumber = 0 // FIP: if such a sector cannot be found, assert that the deal's end epoch has passed and use sector ID 0
```
anorth, line 490: This isn't making the assertion that the FIP specifies. If there's a good reason for that, can it be written in the comment? Should the FIP be changed?

builtin/v13/migration/market.go
anorth: Isn't this the case annotated "// FIP: if such a sector cannot be found, assert that the deal's end epoch has passed and use sector ID 0" in the FromScratch case (in which case yes)?

 Adding the FIP spec as comments in this version too would be helpful. Though this code isn't setting the deal state sector number.

anorth (commented):
The case annotated "// todo is this normal and possible??" seems to be returning an error where it shouldn't.

The FromScratch version looks good but is lacking an assertion that the FIP specifies.

The complexity of the Diff version makes it much harder to verify. I didn't see anything wrong from a FIP specification point of view, but I don't understand the diff mechanics in depth.

builtin/v13/migration/market.go
magik6k: I believe the assert is caught by this invariant

```
acc.Require((deal.SectorNumber == sectorDeal.SectorNumber) || (deal.SectorNumber == 0 && deal.SlashEpoch != -1),
			"deal sector number %d does not match sector %d for miner %v (ds: %#v; ss %#v)",
```

Will add a comment here

builtin/v13/migration/market.go
magik6k: This function is called inside of a `oldState.SlashEpoch != -1` so I believe that this isn't this case, there the sector number will correctly be set to zero

```
		case amt.Add: // both in case of chain going forward and chain reorg this is just new deals
			var oldState market12.DealState
			if err := oldState.UnmarshalCBOR(bytes.NewReader(change.After.Raw)); err != nil {
				return cid.Cid{}, cid.Cid{}, xerrors.Errorf("failed to unmarshal old state: %w", err)
			}

			var newState market13.DealState
			newState.SlashEpoch = oldState.SlashEpoch
			newState.LastUpdatedEpoch = oldState.LastUpdatedEpoch
			newState.SectorStartEpoch = oldState.SectorStartEpoch
			newState.SectorNumber = 0 // terminated / not found (?)

			if oldState.SlashEpoch == -1 {
				si, err := addProviderSectorEntry(deal)
```

arajasek:
Closing in favour of #246

PR #237, opened 2024-01-08
change the value of duration from endEpoch-startEpoch to sectorExpiry-sectorActivation

firesWu:
if the value `endepoch - startepoch` is used as the `duration`, the calculated initial collateral for real deal may be smaller than the actual value, leading to a failed execution of the provecommit msg execution.

### DIFF
--- a/builtin/v12/market/market_state.go
+++ b/builtin/v12/market/market_state.go
@@ -180,7 +180,7 @@ func validateAndComputeDealWeight(proposals *DealArray, dealIDs []abi.DealID, mi
 
 		// Compute deal weight
 		totalDealSpace += uint64(proposal.PieceSize)
-		dealSpaceTime := DealWeight(proposal)
+		dealSpaceTime := DealWeight(proposal, sectorExpiry, sectorActivation)
 		if proposal.VerifiedDeal {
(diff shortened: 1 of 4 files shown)


rvagg (approved):
This makes sense to me. In actors we calculate `duration` the same way: https://github.com/filecoin-project/builtin-actors/blob/8d957d2901c0f2044417c268f0511324f591cb92/actors/miner/src/lib.rs#L5522-L5534

This code gets used in lotus for `StateMinerPreCommitDepositForPower` here: https://github.com/filecoin-project/lotus/blob/21abfc69fb552b92fa3ce223e863f1d582dec237/node/impl/full/state.go#L1463-L1468

But I'd like someone else who works more closely with this to verify this is correct. @magik6k @LexLuthr ?

rvagg (changes requested):
Going to request a rebase on master and changes to `v14` too. At this stage I think the `v12` changes are unnecessary.

rvagg:
@firesWu we have a 👍 on this from the miner code experts, thanks for catching it. Would you mind rebasing and switching your v12 change to v14?

rvagg:
It was easy enough to do in a new PR @ https://github.com/filecoin-project/go-state-types/pull/291

rvagg:
merged as #291, thanks @firesWu

PR #242, opened 2024-02-06
v13: make AggregateProofType optional on ProveCommitSectors3Params & ProveReplicaUpdates3Params

rvagg:
Ref: https://github.com/filecoin-project/FIPs/pull/934

Depends on:

* https://github.com/whyrusleeping/cbor-gen/pull/93
* ~https://github.com/filecoin-project/go-hamt-ipld/pull/117~
* ~https://github.com/filecoin-project/go-amt-ipld/pull/87~

~But I think that https://github.com/whyrusleeping/cbor-gen/pull/92 is going to cause us grief so we may have to deal with that before we move ahead here. It breaks everything that uses cbor-gen unless updated and in lotus we depend on older versions of the hamt and amt via specs-actors so this change here doesn't even work in lotus.~

_Update: https://github.com/whyrusleeping/cbor-gen/pull/93 merged along with https://github.com/whyrusleeping/cbor-gen/pull/94 which solves the const->var dramas. This is good to go._


rvagg:
Updated using a version of cbor-gen that combines both https://github.com/whyrusleeping/cbor-gen/pull/94 and https://github.com/whyrusleeping/cbor-gen/pull/93; so we get backward compatibility and don't need the hamt and amt changes and we get nullable ints.

CI will fail until those are merged, mainly for the `check-gen` but also there's still a need to use a `gen.ReadStringWithMax` call out to cbor-gen which is missing.

Stebalien (approved):

magik6k (approved):

PR #244, opened 2024-02-13
feat: FIP-0076: add new methods and types

arajasek:
This PR adds the new types and methods from FIP-0076, in conformance with corresponding builtin-actors code. It does NOT update the v13 migration to actually migrate to these new types, nor does it add any new invariants.


builtin/v13/market/invariants.go
```diff
@@ -157,10 +157,6 @@ func CheckStateInvariants(st *State, store adt.Store, balance abi.TokenAmount, c
 
 			dealStateCount++
 
-			if dealState.VerifiedClaim != verifreg.NoAllocationID {
```
arajasek, line 160: This is dropped, since deal states no longer record verified claims. New invariants will be added in a future PR.

Stebalien (approved):

PR #246, opened 2024-02-13
feat: FIP-0076: implement migration to populate ProviderSectors

arajasek:
Extracted from #236.

This is a direct copy of the code from there (except for changes made in earlier PRs #245 and #244). I have not yet examined the code for correctness, nor have I addressed some outstanding review on that PR, and so am opening this as draft for now. It effectively needs review including from me.

In particular, reviewers please examine whether the use of the migration cache is incorrect if the "pre-migrated" state is NOT an ancestor of the migration input state. It SHOULD be the case that this is ancestry-agnostic.

### DIFF
--- a/builtin/v13/migration/market.go
+++ b/builtin/v13/migration/market.go
@@ -1,0 +1,409 @@
+package migration
+
+import (
+	"bytes"
+	"context"
+	"errors"
+
+	"github.com/filecoin-project/go-amt-ipld/v4"
+	"github.com/ipfs/go-cid"
+	cbor "github.com/ipfs/go-ipld-cbor"
+	typegen "github.com/whyrusleeping/cbor-gen"
+	"golang.org/x/xerrors"
+
+	"github.com/filecoin-project/go-state-types/abi"
+	market12 "github.com/filecoin-project/go-state-types/builtin/v12/market"
+	market13 "github.com/filecoin-project/go-state-types/builtin/v13/market"
+	miner13 "github.com/filecoin-project/go-state-types/builtin/v13/miner"
+	adt13 "github.com/filecoin-project/go-state-types/builtin/v13/util/adt"
+	"github.com/filecoin-project/go-state-types/builtin/v9/util/adt"
+	"github.com/filecoin-project/go-state-types/migration"
+)
+
+var errItemFound = errors.New("item found")
+
+type marketMigrator struct {
+	providerSectors *providerSectors
+	OutCodeCID      cid.Cid
+}
+
+func newMarketMigrator(ctx context.Context, store cbor.IpldStore, outCode cid.Cid, ps *providerSectors) (*marketMigrator, error) {
+	return &marketMigrator{
+		providerSectors: ps,
+		OutCodeCID:      outCode,
+	}, nil
+}
+
+func (m *marketMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, in migration.ActorMigrationInput) (result *migration.ActorMigrationResult, err error) {
+	var inState market12.State
+	if err := store.Get(ctx, in.Head, &inState); err != nil {
+		return nil, xerrors.Errorf("failed to load market state for %s: %w", in.Address, err)
+	}
+
+	providerSectors, newStates, err := m.migrateProviderSectorsAndStates(ctx, store, in, inState.States, inState.Proposals)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to migrate provider sectors: %w", err)
+	}
+
+	outState := market13.State{
+		Proposals:                     inState.Proposals,
+		States:                        newStates,
+		PendingProposals:              inState.PendingProposals,
+		EscrowTable:                   inState.EscrowTable,
+		LockedTable:                   inState.LockedTable,
+		NextID:                        inState.NextID,
+		DealOpsByEpoch:                inState.DealOpsByEpoch,
+		LastCron:                      inState.LastCron,
+		TotalClientLockedCollateral:   inState.TotalClientLockedCollateral,
+		TotalProviderLockedCollateral: inState.TotalProviderLockedCollateral,
+		TotalClientStorageFee:         inState.TotalClientStorageFee,
+		PendingDealAllocationIds:      inState.PendingDealAllocationIds,
+		ProviderSectors:               providerSectors,
+	}
+
+	newHead, err := store.Put(ctx, &outState)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to put new state: %w", err)
+	}
+
+	return &migration.ActorMigrationResult{
+		NewCodeCID: m.MigratedCodeCID(),
+		NewHead:    newHead,
+	}, nil
+}
+
+func (m *marketMigrator) MigratedCodeCID() cid.Cid {
+	return m.OutCodeCID
+}
+
+func (m *marketMigrator) migrateProviderSectorsAndStates(ctx context.Context, store cbor.IpldStore, in migration.ActorMigrationInput, states, proposals cid.Cid) (cid.Cid, cid.Cid, error) {
+	// providerSectorsRoot: HAMT[ActorID]HAMT[SectorNumber]SectorDealIDs
+
+	okIn, prevInStates, err := in.Cache.Read(migration.MarketPrevDealStatesInKey(in.Address))
+	if err != nil {
+		return cid.Undef, cid.Undef, xerrors.Errorf("failed to get previous inRoot from cache: %w", err)
+	}
+
+	okInPr, prevInProposals, err := in.Cache.Read(migration.MarketPrevDealProposalsInKey(in.Address))
+	if err != nil {
+		return cid.Undef, cid.Undef, xerrors.Errorf("failed to get previous inRoot from cache: %w", err)
+	}
+
+	okOut, prevOutStates, err := in.Cache.Read(migration.MarketPrevDealStatesOutKey(in.Address))
+	if err != nil {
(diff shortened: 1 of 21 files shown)


builtin/v13/migration/market.go
```diff
@@ -0,0 +1,619 @@
...
+				sectorDeals, ok := providerSectorsMem[sid.Miner]
+				if !ok {
+					sectorDeals = make(map[abi.SectorNumber][]abi.DealID)
+					providerSectorsMem[sid.Miner] = sectorDeals
+				}
+				sectorDeals[sid.Number] = append(sectorDeals[sid.Number], deal)
+			} else {
+				//fmt.Printf("deal %d not found in providerSectors: %v\n", deal, oldState)
+
+				newState.SectorNumber = 0 // FIP: if such a sector cannot be found, assert that the deal's end epoch has passed and use sector ID 0
```
Stebalien, line 560: redundant?

builtin/v13/migration/market.go
```diff
@@ -0,0 +1,619 @@
...
+
+	if err := in.Cache.Write(migration.MarketPrevProviderSectorsOutKey(in.Address), providerSectorsRoot); err != nil {
+		return cid.Undef, cid.Undef, xerrors.Errorf("failed to write previous outRoot to cache: %w", err)
+	}
+
+	return providerSectorsRoot, newStateArrayRoot, nil
+}
+
+func (m *marketMigrator) migrateProviderSectorsAndStatesWithDiff(ctx context.Context, store cbor.IpldStore, prevInStatesCid, prevOutStatesCid, prevOutProviderSectorsCid, inStatesCid, prevInProposals, inProposals cid.Cid) (cid.Cid, cid.Cid, error) {
+	diffs, err := amt.Diff(ctx, store, store, prevInStatesCid, inStatesCid, amt.UseTreeBitWidth(market12.StatesAmtBitwidth))
```
Stebalien, line 138: So, unfortunately, we're diffing the wrong states here if we care about re-orgs. I.e., it's possible for a re-org to change the sector number to which a deal is assigned without changing anything about the deal itself, and this diff won't catch that because it's diffing the deal states which don't record sector information.

I discussed this with @arajasek and we think the answer is to just not allow reorgs. I.e., don't use the cached migration if its not an ancestor. That'll ensure that neither sector nor deal IDs are reused.

Stebalien: Basically:

1. Wait for some measure of finality (not real finality, but, like, a few epochs) after the pre-migration epoch before running the pre-migration. OR, run the premigration at the specified epoch, but do a lookback (i.e., migrate state from a few epochs ago).
2. In the "migration" struct itself, record the highest pre-migration tipset.
3. When running the actual migration (or, really, any pre-migrations), if there is a pre-migration tipset recorded and it's not an ancestor of the tipset being migrated, discard the cache.

Stebalien: Ugh. Ok, maybe that's not the way. We _do_ diff the sectors anyways. So...

1. When we diff the sectors, we need to record the changes.
2. When we diff the deal states, we need to apply/remove the sector changes.
3. Finally, we need to iterate over the _rest_ of the sector changes, and fix their deal states.

rjan90: @magik6k is the above note something you can look at addressing?

anorth (commented):
I only reviewed FromScratch, which looks ok to me.

builtin/v13/migration/miner.go
```diff
@@ -0,0 +1,199 @@
...
+				found, err := inSectors.Get(change.Key, &sector)
+				if err != nil {
+					return nil, xerrors.Errorf("failed to get sector %d in inSectors: %w", sectorNo, err)
+				}
+
+				if !found {
+					return nil, xerrors.Errorf("didn't find sector %d in inSectors", sectorNo)
+				}
+
+				if len(sector.DealIDs) == 0 {
```
Stebalien, line 132: Modify needs to be treated as remove + add, so we need to run the remove logic from below even in this case.

builtin/v13/migration/miner.go
```diff
@@ -0,0 +1,199 @@
...
+						Number: sectorNo,
+					}
+				}
+
+				sectorDeals, ok := m.providerSectors.updatesToMinerToSectorToDeals[abi.ActorID(mid)]
+				if !ok {
+					sectorDeals = make(map[abi.SectorNumber][]abi.DealID)
+					m.providerSectors.updatesToMinerToSectorToDeals[abi.ActorID(mid)] = sectorDeals
+				}
+				// There can't already be an entry for this sector, so just set (not append)
```
Stebalien, line 150: reorg

builtin/v13/migration/miner.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("failed to get miner ID: %w", err)
+	}
+
+	inSectors, err := adt.AsArray(ctxStore, inState.Sectors, miner12.SectorsAmtBitwidth)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to load sectors array: %w", err)
+	}
+
+	hasCached, prevSectors, err := in.Cache.Read(migration.MinerPrevSectorsInKey(in.Address))
```
arajasek, line 66: We need to think about use of the cache -- providerSectors is populated based on whether or not there's a cache hit, but the market migration doesn't consider the possibility that some miners were cached, but others weren't.

builtin/v13/migration/market.go
```diff
@@ -0,0 +1,409 @@
...
+			return errItemFound
+		})
+		nonEmpty := false
+		if err == errItemFound {
+			nonEmpty = true
+			err = nil
+		}
+		if err != nil {
+			return cid.Undef, cid.Undef, xerrors.Errorf("failed to iterate actor sectors: %w", err)
+		}
```
Stebalien, line 277: Consider attaching this to the "map" definition in `util/adt/map.go` so I never have to look at this code again?

builtin/v13/migration/market.go
```diff
@@ -0,0 +1,409 @@
...
+
+	for miner, sectors := range m.providerSectors.minerToSectorToDeals {
+		actorSectors, err := adt.MakeEmptyMap(ctxStore, market13.ProviderSectorsHamtBitwidth)
+		if err != nil {
+			return cid.Undef, cid.Undef, xerrors.Errorf("failed to create empty map: %w", err)
+		}
+
+		var sectorDealIDs market13.SectorDealIDs
+		for sector, deals := range sectors {
+			sectorDealIDs = deals
```
Stebalien, line 385: I can't wait for the fixed loop to ship by default. This is one of the dumbest go design decisions.

builtin/v13/migration/miner.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		return nil, xerrors.Errorf("failed to read prev sectors from cache: %w", err)
+	}
+
+	var sector miner.SectorOnChainInfo
+	if !hasCached {
+		// no cached migration, so we simply iterate all sectors and collect deal IDs
+
+		err = inSectors.ForEach(&sector, func(i int64) error {
+			if len(sector.DealIDs) == 0 || sector.Expiration < UpgradeHeight {
```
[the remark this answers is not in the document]
arajasek: Yup, I have a note for that too.

builtin/v13/migration/market.go
```diff
@@ -0,0 +1,409 @@
...
+			}
+
+			newState := market13.DealState{
+				SectorNumber:     0,
+				SectorStartEpoch: oldState.SectorStartEpoch,
+				LastUpdatedEpoch: oldState.LastUpdatedEpoch,
+				SlashEpoch:       oldState.LastUpdatedEpoch,
+			}
+
+			// TODO: We should technically only set this if the deal hasn't expired,
```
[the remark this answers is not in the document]
arajasek: That involves looking up the proposal for every single deal in here, which is significant work, I think.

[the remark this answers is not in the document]
Stebalien: Ah. We don't include that info in the deal state.

What if we check the current epoch is more than `DealMinDuration` after `SectorStartEpoch`. If so, we lookup the deal state.

Honestly, I think looking up the deal state each time won't be a huge issue either.

You're right that this _shouldn't_ matter, it's just a bit annoying that we're relying on the time between the migration and the first pre-migration being short.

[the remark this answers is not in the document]
arajasek: I'm gonna track this as a future item, and see if looking it up each time isn't too much slower.

builtin/v13/migration/market.go
```diff
@@ -0,0 +1,409 @@
...
+
+			if err := prevOutStates.Set(uint64(deal), &newState); err != nil {
+				return cid.Undef, cid.Undef, xerrors.Errorf("failed to set new state: %w", err)
+			}
+		}
+	}
+
+	// now, process updatesToMinerToSectorToDeals, making the appropriate updates
+
+	for miner, sectorUpdates := range m.providerSectors.updatesToMinerToSectorToDeals {
```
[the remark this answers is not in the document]
arajasek: I agree, but I'd like this to be a future item too. We can unblock "having an nv22 network" for now.

Stebalien (approved):
LGTM. This PR is correct (in practice) assuming we detect and avoid using the cache for reorgs in lotus. That means:

1. Performing the premigration at some small lookback from the actual premigration epoch.
2. Recording the tipset where we did the premigration.
3. During the real migration (or any subsequent premigrations), check if the current epoch is a child of the premigration. If not, discard the cache.
4. During the real migration, DO NOT update the cache. If we reorg around the migration epoch, we just need to re-run the migration based on the last premigration.

PR #247, opened 2024-02-14
feat: add names for "common" errors

Stebalien:
That way we get nicer error messages.


exitcode/common.go
```diff
@@ -25,6 +25,8 @@ const (
 	// ErrReadOnly indicates that the actor cannot perform the requested operation
 	// in read-only mode.
 	ErrReadOnly
+	/// The method cannot handle a transfer of value.
```
rvagg, line 28:
```suggestion
	/// ErrNotPayable indicates the method cannot handle a transfer of value.
```

rvagg (approved):
with doc nit

Issue #239, opened 2024-01-22
FIP-0074: Add new types


arajasek:
In order to support FIP-0074 in v13, we need to add the following types:

- SettleDealPaymentsParams 
- SettleDealPaymentsReturn
- DealSettlementSummary 

We also need to add the new `SettleDealPayments` method to the v13 market actor's methods list.

For more, see https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0074.md.

PR #248, opened 2024-02-20
feat: add SettleDealPayments method and types

LexLuthr:
Fixes https://github.com/filecoin-project/go-state-types/issues/239

### DIFF
--- a/builtin/v13/market/market_types.go
+++ b/builtin/v13/market/market_types.go
@@ -140,18 +140,16 @@ type GetDealSectorParams = DealQueryParams
 
 type GetDealSectorReturn = abi.SectorNumber
 
-type SettleDealPaymentsParams struct {
-	Deals bitfield.BitField
-}
+type SettleDealPaymentsParams = bitfield.BitField
 
-type SettleDealPaymentsReturnResultFailCodes struct {
+type FailCode struct {
 	Index    uint64
 	ExitCode exitcode.ExitCode
 }
 
-type SettleDealPaymentsReturnResult struct {
+type BatchReturn struct {
 	SuccessCount uint64
-	FailCount    []SettleDealPaymentsReturnResultFailCodes
+	FailCount    []FailCode
 }
 
 type DealSettlementSummary struct {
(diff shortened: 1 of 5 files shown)


builtin/v13/market/market_types.go
```diff
@@ -138,3 +139,31 @@ type GetDealActivationReturn struct {
 type GetDealSectorParams = DealQueryParams
 
 type GetDealSectorReturn = abi.SectorNumber
+
+type SettleDealPaymentsParams struct {
+	Deals bitfield.BitField
```
Stebalien, line 144: Should be:

```go
type SettleDealPaymentParams = bitfield.BitField
```

(the upstream type "transparently" encodes as a bitfield without wrapping it in a struct)

builtin/v13/market/market_types.go
```diff
@@ -138,3 +139,31 @@ type GetDealActivationReturn struct {
...
+}
+
+type SettleDealPaymentsReturnResultFailCodes struct {
```
Stebalien, line 147: You can just call this a `FailCode`.

Stebalien: (we may want to move it to a shared location later anyways)

builtin/v13/market/market_types.go
```diff
@@ -138,3 +139,31 @@ type GetDealActivationReturn struct {
...
+	Index    uint64
+	ExitCode exitcode.ExitCode
+}
+
+type SettleDealPaymentsReturnResult struct {
```
Stebalien, line 152: `BatchReturn`

builtin/v13/market/market_types.go
Stebalien: Yeah, we'll move both `FailCode` and `BatchReturn` to `util` in a followup patch (removing it from the verifreg.

LexLuthr:
@Stebalien I have updated the names as per the suggestions. I have not defined `market.SettleDealPaymentsParams{}` in gen.go since it is unwrapped and bitfield.Bitfield already has Marshal and Unmarshal methods.

Stebalien (approved):

PR #260, opened 2024-04-19
Experimental market migration memory reduction

ZenGround0:
After experiencing OOM kills @rvagg shared pprof report showing ~80G of memory held by the market migration.  

pprof shows that the culprit appears to be AMT Get reading in market proposals which are the market actor's state bottleneck.  

I haven't been able to understand how memory footprint could be so high. All of f05's state is ~10GB and AMTs should always be caching intermediate nodes so there shouldn't be extra reads.  However going from 10GB -> 10MB of memory needed with little expected impact on speed is already a good thing.


builtin/v13/migration/market.go
```diff
@@ -309,10 +309,21 @@ func (m *marketMigrator) migrateProviderSectorsAndStatesFromScratch(ctx context.
...
+
 	var oldState market12.DealState
 	var newState market13.DealState
 
 	err = oldStateArray.ForEach(&oldState, func(i int64) error {
+		// Refresh proposals array periodically to avoid holding onto all ~10GB of memory
+		// throughout whole migration.
+		// This has limited impact on caching speedups because the access pattern is sequential.
+		if i%(int64(proposalsSize)/1000) == 0 {
+			proposalsArr, err = adt.AsArray(ctxStore, proposals, market12.ProposalsAmtBitwidth)
```
Stebalien, line 322: Don't we need to flush?

Stebalien:
> I haven't been able to understand how memory footprint could be so high. All of f05's state is ~10GB and AMTs should always be caching intermediate nodes so there shouldn't be extra reads. However going from 10GB -> 10MB of memory needed with little expected impact on speed is already a good thing.

So, iteration doesn't cache (in go, IIRC). Random lookups do and I'm guessing our library isn't super optimized for that.

Stebalien:
Oh, wow. Our deferred cbor type is terrible.

builtin/v13/migration/market.go
```diff
@@ -309,10 +309,21 @@ func (m *marketMigrator) migrateProviderSectorsAndStatesFromScratch(ctx context.
...
+	proposalsSize := proposalsArr.Length()
+
 	var oldState market12.DealState
 	var newState market13.DealState
 
 	err = oldStateArray.ForEach(&oldState, func(i int64) error {
+		// Refresh proposals array periodically to avoid holding onto all ~10GB of memory
+		// throughout whole migration.
+		// This has limited impact on caching speedups because the access pattern is sequential.
+		if i%(int64(proposalsSize)/1000) == 0 {
```
rvagg, line 321:
```suggestion
		if proposalsSize > 0 && i%(int64(proposalsSize)/1000) == 0 {
```

picked up @ https://app.circleci.com/pipelines/github/filecoin-project/lotus/33637/workflows/b64e4593-54cb-437f-bbdc-dbe756bd2280/jobs/1210829

builtin/v13/migration/market.go
```diff
@@ -309,10 +309,21 @@ func (m *marketMigrator) migrateProviderSectorsAndStatesFromScratch(ctx context.
...
+	proposalsSize := proposalsArr.Length()
+
 	var oldState market12.DealState
 	var newState market13.DealState
 
 	err = oldStateArray.ForEach(&oldState, func(i int64) error {
+		// Refresh proposals array periodically to avoid holding onto all ~10GB of memory
+		// throughout whole migration.
+		// This has limited impact on caching speedups because the access pattern is sequential.
+		if proposalsSize > 0 && i%(int64(proposalsSize)/1000) == 0 {
```
Stebalien, line 321: Are we sure we don't have holes in this array? I think we want to count instead of using `i`.

ZenGround0: Thx, we will definitely have holes

rvagg (approved):

PR #261, opened 2024-04-22
fix: fix v13 migration panic when proposals < 1000

Stebalien:
If we had fewer than 1000 proposals (only relevant when testing, really), we'd panic because `proposalsSize/1000 == 0` and `cnd % 0` leads to a divide by zero panic.

### DIFF
--- a/builtin/v13/migration/market.go
+++ b/builtin/v13/migration/market.go
@@ -318,8 +318,10 @@ func (m *marketMigrator) migrateProviderSectorsAndStatesFromScratch(ctx context.
 		// Refresh proposals array periodically to avoid holding onto all ~10GB of memory
 		// throughout whole migration.
 		// This has limited impact on caching speedups because the access pattern is sequential.
+		// The minimum batch-size is 100 (hence a 100,000 minimum proposals before this
+		// kicks in).
 		cnt += 1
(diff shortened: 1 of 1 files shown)


builtin/v13/migration/market.go
```diff
@@ -318,8 +318,10 @@ func (m *marketMigrator) migrateProviderSectorsAndStatesFromScratch(ctx context.
 		// Refresh proposals array periodically to avoid holding onto all ~10GB of memory
 		// throughout whole migration.
 		// This has limited impact on caching speedups because the access pattern is sequential.
+		// The minimum batch-size is 100 (hence a 100,000 minimum proposals before this
+		// kicks in).
 		cnt += 1
-		if proposalsSize > 0 && cnt%(int(proposalsSize/1000)) == 0 {
+		if proposalsSize > 100_000 && cnt%(int(proposalsSize/1000)) == 0 {
```
Stebalien, line 324: note: I'm not sure this is how we want to write this anyways. IMO, we want:

```suggestion
		if cnt%K == 0 {
```

where `K` is an experimentally derived number that keeps the memory below some set limit.

(not worth changing now, honestly)

masih (approved):

PR #263, opened 2024-05-07
test: verifreg: add initial cbor encoding forms tests

rvagg:
Since I was touching type changes in https://github.com/filecoin-project/go-state-types/pull/262 I thought I'd do an initial version of something I've proposed doing a more comprehensive form of: cbor encoding and round-trip testing of publicly exported types, which is particularly important for the types we pass through builtin-actors. So for now I'm just starting with  ClaimAllocationsParams, and am following this up with a PR to builtin-actors with identical fixtures.

Ideally we'd expand this, perhaps slowly, to cover the full suite of types that pass through this barrier, and in the process provide fixtures for other implementations to test against as well.

The diagnostic output might be overkill, I can remove that if someone wants to object, I just like being absolutely clear.

### DIFF
--- a/builtin/v13/verifreg/verifreg_types_test.go
+++ b/builtin/v13/verifreg/verifreg_types_test.go
@@ -17,24 +17,11 @@ func TestSerializationSlaimAllocationsParams(t *testing.T) {
 		hex    string
 	}{
 		{
-			/*
-			 82                                                # array(2)
-			   80                                              #   array(0)
-			   f4                                              #   false
-			*/
 			params: ClaimAllocationsParams{Sectors: nil, AllOrNothing: false},
-			hex:    "8280f4",
+			// [[],false]
+			hex: "8280f4",
 		},
 		{
-			/*
-			 82                                                # array(2)
-			   81                                              #   array(1)
-			     83                                            #     array(3)
-			       18 65                                       #       uint(101)
-			       18 ca                                       #       uint(202)
-			       80                                          #       array(0)
-			   f5                                              #   true
-			*/
 			params: ClaimAllocationsParams{
 				Sectors: []SectorAllocationClaims{{
 					Sector:       101,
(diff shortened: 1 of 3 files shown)


rvagg:
Rust version: https://github.com/filecoin-project/builtin-actors/pull/1538

rvagg:
removed diagnostic output, the more I look at it the more wasteful it seems when we expand the number of these

builtin/v13/verifreg/verifreg_types_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"encoding/hex"
+	"testing"
+
+	"github.com/ipfs/go-cid"
+	"github.com/stretchr/testify/require"
+)
+
+// Tests to match with Rust fil_actor_verifreg::serialization
+
+func TestSerializationSlaimAllocationsParams(t *testing.T) {
```
ZenGround0, line 14: nit: ClaimAllocations for all of these

ZenGround0 (approved):
It might be overkill to do this for multiple versions of a type when nothing changes.

If we could remove these from next to the code and put them in some place where all versions are accessible we could maybe only update tests when the actual types change.  That would reduce code volume a lot for the same value.  Idk though it's pretty cheap to copy paste a test and carry all these along.

builtin/v13/verifreg/verifreg_types_test.go
rvagg: embarrassing

rvagg:
Regarding duplication - my theory here is that as we start to accumulate these, a pattern should emerge that can be extracted to make this simpler.

Perhaps reviving https://github.com/filecoin-project/test-vectors at some point for these would be good and then run the tests from both ends. For IPLD I set up this: https://github.com/ipld/codec-fixtures which runs a set of fixtures across the latest codec implementations of Go, JS and Rust each night and complains if there are problems. I think a similar pattern could work here, but I don't really want to dive in too deep right now.

PR #269, opened 2024-05-23
Feat: NI-PoRep

rjan90:
PR for enabling NI-PoRep in GST.


abi/sector.go
```diff
@@ -284,6 +298,41 @@ var SealProofInfos = map[RegisteredSealProof]*SealProofInfo{
 		WindowPoStProof:  RegisteredPoStProof_StackedDrgWindow64GiBV1,
 		UpdateProof:      RegisteredUpdateProof_StackedDrg64GiBV1,
 	},
+	RegisteredSealProof_StackedDrg2KiBV1_1_Feat_NiPoRep: {
+		ProofSize:        192,
```
rvagg, line 302: we'd better get these right before merging this

abi/sector.go
```diff
@@ -91,6 +91,12 @@ const (
 	RegisteredSealProof_StackedDrg512MiBV1_1_Feat_SyntheticPoRep = RegisteredSealProof(12)
 	RegisteredSealProof_StackedDrg32GiBV1_1_Feat_SyntheticPoRep  = RegisteredSealProof(13)
 	RegisteredSealProof_StackedDrg64GiBV1_1_Feat_SyntheticPoRep  = RegisteredSealProof(14)
+
+	RegisteredSealProof_StackedDrg2KiBV1_1_Feat_NiPoRep   = RegisteredSealProof(15)
```
rvagg, line 95: @vmx should we be naming these `1_2` instead of `1_1`, I know there's a bit of internal confusion but what is the _correct_ way to name them since what we do here sets the tone for the rest of the Go stack.

vmx: For consistency with the Rust side it would make sense to name them `1_2`.

rjan90: Renamed to V1_2 in https://github.com/filecoin-project/go-state-types/pull/269/commits/0c8092c05dd9b752ba70ee82ca79fec63186611d

abi/sector.go
rjan90: Updated the proofs-sizes here: https://github.com/filecoin-project/go-state-types/pull/269/commits/2763495cb7447abc95bb54059ff3cd11f07cbfde

rvagg (approved):

PR #270, opened 2024-05-24
feat: miner: niporep params and cbor encoding vector tests

rvagg:
Params implementation and tests with vectors to match https://github.com/filecoin-project/go-state-types/pull/270

### DIFF
--- a/builtin/v14/miner/miner_types.go
+++ b/builtin/v14/miner/miner_types.go
@@ -502,7 +502,7 @@ type SectorNIActivationInfo struct {
 	SealingNumber abi.SectorNumber // Sector number used to generate replica id
 	SealerID      abi.ActorID      // Must be set to ID of receiving actor for now
 	SealedCID     cid.Cid          // CommR
-	SectorNumber  abi.SectorNumber // unique id of sector in actor state
+	SectorNumber  abi.SectorNumber // Unique id of sector in actor state
 	SealRandEpoch abi.ChainEpoch
 	Expiration    abi.ChainEpoch
 }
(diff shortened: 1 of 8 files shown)


builtin/v14/miner/policy.go
```diff
@@ -107,6 +107,16 @@ var MaxPreCommitRandomnessLookback = builtin.EpochsInDay + ChainFinality // PARA
...
 
+// The maximum number of sector commitments in a single batch for calling ProveCommitSectorsNI.
+// Same as PreCommitSectorBatchMaxSize for consistency.
+var ProveCommitNiSectorBatchMaxSize = 256 // PARAM_SPEC
+
+// Maximum number of epochs within which to fetch a valid seal randomness from the chain for
+// a non-interactive PoRep proof. This balances the need to tie the seal to a particular chain with
+// but makes allowance for service providers to offer pre-sealed sectors within a larger window of
+// time.
+var MaxProveCommitNiLookback = abi.ChainEpoch(30*builtin.EpochsInDay) + ChainFinality // PARAM_SPEC
```
rvagg, line 118: this will likely become 180 after a fip change

builtin/v14/miner/miner_types.go
```diff
@@ -496,3 +496,32 @@ type SectorReturn = []PieceReturn
...
+	SealerID      abi.ActorID      // Must be set to ID of receiving actor for now
+	SealedCID     cid.Cid          // CommR
+	SectorNumber  abi.SectorNumber // unique id of sector in actor state
+	SealRandEpoch abi.ChainEpoch
+	Expiration    abi.ChainEpoch
+}
+
+// ProveCommitSectorsNIParams is the parameters for non-interactive prove committing of sectors
+// via the miner actor method ProveCommitSectorsNI.
+type ProveCommitSectorsNIParams struct {
```
rjan90, line 512: We probably want to add `ProvingDeadline uint64` eventually in here as well. Given that we want to land https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0092.md together with NI-PoRep.

But maybe worth waiting until that has moved to Last Call

builtin/v14/miner/policy.go
rjan90:
```suggestion
var MaxProveCommitNiLookback = abi.ChainEpoch(180*builtin.EpochsInDay) + ChainFinality // PARAM_SPEC
```

rvagg: need to remove finality too

rjan90: Ah, correct! Updated the suggested change.

rvagg: done

builtin/v14/miner/policy.go
```diff
@@ -107,6 +107,16 @@ var MaxPreCommitRandomnessLookback = builtin.EpochsInDay + ChainFinality // PARA
 // (2) prevents a miner attempting a long fork in the past to insert a pre-commitment after seeing the challenge.
 var PreCommitChallengeDelay = abi.ChainEpoch(150) // PARAM_SPEC
 
+// The maximum number of sector commitments in a single batch for calling ProveCommitSectorsNI.
+// Same as PreCommitSectorBatchMaxSize for consistency.
+var ProveCommitNiSectorBatchMaxSize = 256 // PARAM_SPEC
```
rvagg, line 112: ugh, this needs to go too

PR #276, opened 2024-06-21
Update CSPV naming

aarshkshah1992:
This reflects the changes made in https://github.com/filecoin-project/builtin-actors/pull/1553


rvagg (commented):
This seems fine. I'm out of the loop on naming so I'll take your word for it. But I believe this needs to be against feat/nv23 instead of master.

rvagg (approved):

ZenGround0 (approved):

PR #279, opened 2024-06-27
feat: rename Actor.Address to Actor.DelegatedAddress

Stebalien:
See https://github.com/filecoin-project/lotus/pull/12155

### DIFF
--- a/builtin/v14/migration/top.go
+++ b/builtin/v14/migration/top.go
@@ -127,9 +127,9 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID 
 			Balance:    f090OldAct.Balance,
 
 			// changed
-			Code:    newAccountCodeCID,
-			Head:    h,
-			Address: &f090ID,
+			Code:             newAccountCodeCID,
+			Head:             h,
+			DelegatedAddress: &f090ID,
 		})
 	}
 	if err := f090Migration(actorsOut); err != nil {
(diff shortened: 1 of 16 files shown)


builtin/v14/migration/top.go
```diff
@@ -127,9 +127,9 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
 			Balance:    f090OldAct.Balance,
 
 			// changed
-			Code:    newAccountCodeCID,
-			Head:    h,
-			Address: &f090ID,
+			Code:             newAccountCodeCID,
+			Head:             h,
+			DelegatedAddress: &f090ID,
```
Stebalien, line 132: @ZenGround0 this still needs to be removed (and sorry about this, we should have fixed this ages ago).

rvagg (approved):
ok, but you could just revert https://github.com/filecoin-project/go-state-types/commit/dfbb8139ee07ab4a0b0abf8ac72e30b465c12f96 while you're in here and be done with it

builtin/v14/migration/top.go
ZenGround0: no problemo I'll file a PR

ZenGround0 (approved):

ZenGround0:
rather than revert I'll explicitly set to nil to handle the case where f090 is an account actor that got us looking into all this

PR #284, opened 2024-07-02
Fix vet and staticcheck errors

rvagg:
Ref: https://github.com/filecoin-project/go-state-types/pull/283

Couldn't resist when I saw the failures there. I'd really like unified-ci to be working in this repo so here's a start. Doesn't fix the test race but deals with the biggest set of problems.


rvagg:
This is in #283, probably doesn't need to be accepted and merged into master and can be closed after #283 is done.

ZenGround0:
Closing since #283 is in.

PR #286, opened 2024-07-03
feat!: extract BatchReturn to batch package, add methods & tests

rvagg:
I started off by wanting methods on `BatchReturn` to roughly match what's available in Rust, then I remembered that I was going to add serialization tests for this too because it's in the NI-PoRep path, then I realised that this struct is used in multiple actors and hasn't changed form since it was introduced. In builtin-actors it's in the shared "fil_actors_runtime" package for this very reason. So this PR introduces the same thing here:

* Extract `BatchReturn` and `FailCode` into a new "batch" package
* Remove all copies, through all actor versions
* Add methods to `BatchReturn`: `Size()`, `AllOk()`, `Codes() []exitcode.ExitCode`, `CodeAt(n uint64) exitcode.ExitCode`
* Add tests for new methods
* Add tests for serializations

This is a **breaking change** because it removes types that were previously there. There are no current direct uses of this type within Lotus today (I'm in the process of introducing one though). This would be a good candidate to ship in a v0.14.0 final I think.


BigLep:
> This would be a good candidate to ship in a v0.14.0 final I think.

For now, I added a note to https://github.com/filecoin-project/lotus/issues/12109 as this being an item for the "Mainnet Upgrade Release"

BigLep:
2024-07-11 maintainer conversation: this will get merged after the nv24 release.

batch/batch_return.go
```diff
@@ -0,0 +1,48 @@
...
+func (b BatchReturn) AllOk() bool {
+	return len(b.FailCodes) == 0
+}
+
+func (b BatchReturn) Codes() []exitcode.ExitCode {
+	codes := make([]exitcode.ExitCode, b.Size())
+	i := 0
+	for _, fc := range b.FailCodes {
+		if fc.Idx > uint64(i) {
+			for ; i < int(fc.Idx); i++ {
```
ZenGround0, line 28: Here we use the invariant that FailCodes are in ascending order.  This is true for all BatchReturn constructed using BatchReturnGen which should be everything on chain.  I think this is a good assumption for this code.

I think we should consistently use this same invariant during traversal inside of CodeAt to stop traversal and return exitcode.OK if fc.Idx > n.

ZenGround0: Alternatively since batchreturn has a constant number of existcodes it might be overall better to use the less efficient but simple 

```golang

func (b BatchReturn) CodeAt(n uint64) exitcode.ExitCode {
     return b.Codes(n)
}
```

batch/batch_return.go
```diff
@@ -0,0 +1,48 @@
...
+		i++
+	}
+	for ; i < len(codes); i++ {
+		codes[i] = exitcode.Ok
+	}
+	return codes
+}
+
+func (b BatchReturn) CodeAt(n uint64) exitcode.ExitCode {
+	for _, fc := range b.FailCodes {
```
ZenGround0, line 42: We probably want bounds checking no matter how we implement this.  Right now the out of bounds case leads to exitcode.Ok which is not the best way to fail.

batch/batch_return.go
```diff
@@ -0,0 +1,48 @@
+package batch
+
+import "github.com/filecoin-project/go-state-types/exitcode"
+
+type BatchReturn struct {
```
ZenGround0, line 5: Maybe add a validate check to the interface? This way we can be sure the data we're working with is good.  Two checks I see are that we're getting monotonic increasing fail codes and all gaps between fail codes are covered by success count.

ZenGround0 (approved):
I had no idea how bad this had gotten.  Thanks for cleaning up

batch/batch_return.go
rvagg: @ZenGround0 do you have an opinion on maximum size of this thing? Can we max out at `math.MaxInt32` so we don't end up making supidly big slices? (even that's stupid-big but ..)

rvagg:
> after the nv24 release

I just need to get back to this and get it landed in the next major release .. so :soon:

ZenGround0 (approved):

PR #287, opened 2024-07-05
chore: Cleanup invariant errors related to v13 migration

rjan90:
I encountered both these errors when running the invariant-check on mainnet. Deleting these two invariants given that they existed only to test the v13 migration as per the TODO in code comments.

- `"MIGRATION-ONLY: un-slashed deal %d not referenced in active sectors of miner %v", dealID, deal.Provider"`
- `"MIGRATION-ONLY: deal %d not found in miner %v for sector %v", dealID, maddr, sectorID)"`

### DIFF
--- a/builtin/v13/check.go
+++ b/builtin/v13/check.go
@@ -267,11 +267,8 @@ func CheckDealStatesAgainstSectors(acc *builtin.MessageAccumulator, minerSummari
 			continue
 		}
 
-		// TODO: If you are reading this error after nv22 (v13 actors), delete this invariant.
-		// It exists to test the v13 migration ONLY.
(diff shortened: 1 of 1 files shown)


rvagg (approved):
I had to go remind myself why this was OK to skip now. As per 0076:

> The per-sector `DealIDs` field is deprecated and to be ignored.
> The miner actor will not write new deals to it.

Which is why we're accumulating them now.

But also, I think this is going to be irrelevant after nv23 because we won't be executign this code path once v14/check.go takes over. I guess this will take effect once we ship 1.28.0 for the short period before nv23?

rjan90:
Rebase on master, so we could merge this.

PR #288, opened 2024-07-08
chore: include manifest & proof in `make gen`; add extra proof types

rvagg:
The `manifest` line in Makefile is included in #283 because it's needed to get past new linting. But `proof` is not included there and I wanted to serialise a couple of the Aggregate types in there for debugging but they don't yet have cbor-gen additions. We don't strictly need these for communication between lotus and actors, but they are handy to have for debugging!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,7 @@ tidy:
 gen:
 	$(GO_BIN) run ./gen/gen.go
 	$(GO_BIN) run ./manifest/gen/gen.go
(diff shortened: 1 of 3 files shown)


aarshkshah1992 (approved):

aarshkshah1992:
@rvagg This looks straight forward. Please merge when ready.

ZenGround0 (approved):

PR #297, opened 2024-08-02
Update all dependencies, add dependabot config

rvagg:


rvagg:
hah, why is staticcheck deciding to be difficult on this one with xerrors.Is?

ZenGround0 (approved):

Issue #285, opened 2024-07-03
Optizmize base migrations


rjan90:
When benchmarking the nv23 migration on my mainnet node, I get these bench-times:

Migration without cache is taking 41seconds:
```
migration height  4056425
old cid  bafy2bzaceafswlgag4nowendbxzwxvvt4ofn4vyp75kbmdods4x2vbsepgt5q
new cid  bafy2bzacecparpmzboywlcpuqz5vp2sn47lk4d3bc24oksvynlbfs4v7vngwy
completed round actual (without cache), took  41.730056334s
manifest cid: bafy2bzacecbueuzsropvqawsri27owo7isa5gp2qtluhrfsto2qg7wpgxnkba
```

Migration with cache is taking 36 seconds which is higher then what our threshold for migration times - which should be less then 1 epoch (30s)
```
completed round actual (with cache), took  36.853138672s
```

- These high times are unexpected given that the migration for network upgrade is very light.
- Resource usage when running the migration on this node is low, its consuming around 10GiB.

ZenGround0:
All right here's a high level summary of time spent.  For more precise understanding of the code corresponding to each log line check out the timing branch here: https://github.com/filecoin-project/go-state-types/compare/master...spike/timing-migration
Timing without cache
```
./lotus-shed migrate-state --repo=/mnt/nvmeraid0/daemon/ 23 bafy2bzacedgsnt7hkqwisxxfcndalrf4claltpyel6k6hixqjyhikg53puq6u
2024-07-03T19:03:27.924Z	WARN	fil-consensus	filcns/upgrades.go:2743	time to initialize migration: 461.255µs
2024-07-03T19:03:49.019Z	WARN	fil-consensus	filcns/upgrades.go:2743	time to run migration: 21.095439755s
2024-07-03T19:03:49.019Z	WARN	fil-consensus	filcns/upgrades.go:2743	time to run f090 migration: 44.545µs
2024-07-03T19:03:51.295Z	WARN	fil-consensus	filcns/upgrades.go:2743	time to flush actorsOut: 2.275956678s
completed round actual (without cache), took  24.891692657s
```

Timing with cache 
```
2024-07-03T19:03:52.941Z	WARN	fil-consensus	filcns/upgrades.go:2743	time to initialize migration: 342.139µs
2024-07-03T19:04:24.281Z	WARN	fil-consensus	filcns/upgrades.go:2743	time to run migration: 31.339779092s
[...]

PR #298, opened 2024-08-03
HAMTv4 using generics instead of cbg.Deferred for perf gains

rvagg:
Ref: https://github.com/filecoin-project/go-state-types/issues/285

Continuing on the optimisation journey with the belief that allocations and GC are taking up a significant portion of the time, I  experimented with removing the `cbg.Deferred` use in go-hamt-ipld and using Go generics to replicate that functionality. When the HAMT doesn't know what value type it's got stored, it skips over the CBOR tokens until it finds a whole unit (single scalar token, whole nested object, whatever) and then slices the bytes for that value and passes it on in `cbg.Deferred` for the caller to decode themselves. This means double CBOR parsing and the allocation and GC of a byte slice that really shouldn't be there. So instead we can pass our serializable type down into the HAMT and it can just defer directly to it to decode its values. That work is in https://github.com/filecoin-project/go-hamt-ipld/pull/122 and this branch depends on that branch as `github.com/filecoin-project/go-hamt-ipld/v4`, keeping `/v3` intact.

Summary of changes so far:

* `builtin.ActorTree` is now an interface which only has the `ActorsV5` method variants and I've also made `builtin.LegacyActorTree` which has those plus the `ActorsV4` method variants.
* `builtin.NewLegacyTree` and `builtin.LoadLegacyTree` are the new names for the original form, these will give you what was always returned before.
* `GetStore()` on both interfaces, and `GetMap()` on `LegacyActorTree`—I need the latter for lotus, which only knows about `adt.Map` for now but with this change `adt.Map` can become a legacy concern and we can just use `hamt.Node[*ActorV5]` to get basically the same thing.
* `RunMigration` only takes an `ActorTree`, it doesn't need the v4 actors; it also uses `builtin.NewTree` to make the new v4 hamt with generics variety
* v14 migration uses `builtin.LoadTree` to get the new v4 hamt with generics to pass to `RunMigration`; the other migrations still use `LoadLegacyTree`.

So the impact of this is that when running the v14 migration we read and write the actors HAMT with the new v4 HAMT code that uses generics instead of `cbg.Deferred`.

---

## Results

`lotus-shed migrate-state 23 bafy2bzacecnpvunvyytfzmdofrdwk2jr5sf4cuzit6o6uctrzqdltqwzbwwmk` for both of them:

### Current lotus & go-state-types

migration height  4145482
old cid  bafy2bzacedpcjm7wgoq6ft3x7jdjtjdmbaqil2zjd2pa6exl2mwpv5opik4s2
new cid  bafy2bzacebpjrptlb44oaj4vyaaa5th3lvqwv5cogdfd6agzvjqr223grwhli
completed round actual (without cache), took  11.323758969s
completed round actual (with cache), took  9.755098713s

### This branch and HAMT v4

migration height  4145482
old cid  bafy2bzacedpcjm7wgoq6ft3x7jdjtjdmbaqil2zjd2pa6exl2mwpv5opik4s2
new cid  bafy2bzacebpjrptlb44oaj4vyaaa5th3lvqwv5cogdfd6agzvjqr223grwhli
completed round actual (without cache), took  10.133880646s
completed round actual (with cache), took  8.987781194s

So we're about 10.5% faster without cache and nearly 8% faster with cache.

### Flame graphs

Before:
flame

After:
flame2

You can see in there:

* `ForEach` on the actors tree goes from 30.9% to 28.90% (most of the action is captured under here since it nests the `Set` too
* `Set` is 11.25% vs 6.22% (this is dramatic because we get to defer serialization whereas previously we had to serialize the value into a slice and hold onto it till a flush)
* `Flush` goes up from 4.97% to 6.51% (this is where the value serialization is moved to, but it's much cheaper because we go directly to the `io.Writer` rather than a slice allocation [+GC], hold, then write)

### Profiles

(click to see full profiles)

Before:
migrate-state cpuprofile

After:
migrate-state cpuprofile2

These are some of the bits I was most interested in:

Screenshot 2024-08-03 at 7 14 16 PM
vs
Screenshot 2024-08-03 at 7 14 29 PM

<br>

Screenshot 2024-08-03 at 7 15 54 PM
vs
Screenshot 2024-08-03 at 7 16 08 PM

(GC improvements not as dramatic as I hoped, but still there)

There's probably a few more tweaks that could be done to make this nicer. There's some internal mutation code in the HAMT that I think should be changed now with the generics that may speed up the path through the code for inserts and deletes.


rvagg:
Bailing on this for now, as per comment @ https://github.com/whyrusleeping/cbor-gen/pull/103#issuecomment-2398852941

PR #308, opened 2024-08-19
fix(miner)!: remove DEAL_WEIGHT_MULTIPLIER and its input to QAP calc

rvagg:
Ref: https://github.com/filecoin-project/builtin-actors/issues/1573
Ref: https://github.com/filecoin-project/builtin-actors/pull/1574

Test copied across from Rust.

This will have implications for Lotus because of the breaking API change, it'd probably be worth figuring out how much pain that is before merging this.


Stebalien:
Given that this doesn't actually change the power calculus, the breaking change shouldn't be a huge issue (no need to abstract over different versions, we can just change it everywhere).

Stebalien (approved):

PR #314, opened 2024-09-27
FIP-0081 migration

rvagg:


migration/util.go
```diff
@@ -87,6 +87,10 @@ type Config struct {
 	ProgressLogPeriod time.Duration
 	// The epoch at which the upgrade will run.
 	UpgradeEpoch abi.ChainEpoch
+
+	// FIP-0081 constants for the power actor state for pledge calculations.
+	PowerRampStartEpoch     int64  // Epoch at which the new pledge calculation starts.
```
ZenGround0, line 92: Same comment from lotus PR -- please file an issue to clean this up after nv24

rvagg: I decided to move these to function arguments since we can't really remove them as we need the migration to stay in place into the future, so these would get stuck in this struct.

ZenGround0 (approved):

PR #320, opened 2024-10-29
chore: bump version to v0.15.0

rjan90:
chore: bump version to v0.15.0

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,3 +1,3 @@
 {
-  "version": "v0.15.0-rc1"
+  "version": "v0.15.0"
 }


rjan90:
@rvagg looks like we had a couple of dependency bumps since the [cutting the rc1 of v0.15.0](https://github.com/filecoin-project/go-state-types/pull/320#issuecomment-2445207504). Should we instead go for a re-tag of v0.15.0-rc1? Or are these very minimal changes, as you [noted here](https://github.com/filecoin-project/go-state-types/pull/317#issuecomment-2398462097).

rvagg:
@rjan90 it should be fine proceeding with this. All of the updates are fairly trivial and will be overridden by whatever lotus is using anyway. The cbor-gen one didn't involve any codegen changes here even that doesn't touch anything here. But for the same reason, re-tagging v0.15.0-rc1 would essentially leave us with the same code being run in lotus as if we tagged off master. So it's up to you!

rvagg (approved):

PR #329, opened 2024-11-28
fix(migration): improve job progress logging in MigrateStateTree

virajbhartiya:
reference to https://github.com/filecoin-project/lotus/pull/12732
This pull request includes an update to the logging format in the `MigrateStateTree` function to improve clarity during the migration process.

Logging improvements:

* [`builtin/v9/migration/top.go`](diffhunk://#diff-0f71e4e5376421454dd6e1792714840e621766fb31e279de1749157bf59a8bd3L400-R405): Changed the log message to include the percentage of jobs completed and rephrased the message for better readability.

### DIFF
--- a/builtin/v9/migration/top.go
+++ b/builtin/v9/migration/top.go
@@ -395,13 +395,14 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID 
 			for {
 				select {
 				case <-time.After(cfg.ProgressLogPeriod):
-					jobsNow := jobCount // Snapshot values to avoid incorrect-looking arithmetic if they change.
-					doneNow := doneCount
-					pendingNow := jobsNow - doneNow
+					jobsNow := atomic.LoadUint32(&jobCount) // Snapshot values to avoid incorrect-looking arithmetic if they change.
+					doneNow := atomic.LoadUint32(&doneCount)
 					elapsed := time.Since(startTime)
 					rate := float64(doneNow) / elapsed.Seconds()
-					log.Log(rt.INFO, "%d jobs created, %d done, %d pending after %v (%.0f/s)",
-						jobsNow, doneNow, pendingNow, elapsed, rate)
+					percentComplete := float64(doneNow) / float64(jobsNow) * 100
+
+					log.Log(rt.INFO, "Performing migration: %d of %d jobs complete (%.1f%%, %.0f/s)",
+						doneNow, jobsNow, percentComplete, rate)
 				case <-workersFinished:
 					return
 				case <-ctx.Done():
--- a/builtin/v9/migration/top.go
+++ b/builtin/v9/migration/top.go
@@ -397,12 +397,11 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID 
 				case <-time.After(cfg.ProgressLogPeriod):
 					jobsNow := jobCount // Snapshot values to avoid incorrect-looking arithmetic if they change.
 					doneNow := doneCount
+					pendingNow := jobsNow - doneNow
 					elapsed := time.Since(startTime)
 					rate := float64(doneNow) / elapsed.Seconds()
-					percentComplete := float64(doneNow) / float64(jobsNow) * 100
-
-					log.Log(rt.INFO, "Performing migration: %d of %d jobs complete (%.1f%%, %.0f/s)",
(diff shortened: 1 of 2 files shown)


rvagg:
@virajbhartiya do this to `builtin/v16/migration/top.go` instead of v9. Actors v14 is for network version 25, the next one (which hasn't been scheduled), so it'll actually be used then. v9 on mainnet was in 2022 for nv17.

After each network upgrade, one of the tasks performed is that someone (usually @rjan90) opens a PR here that copies the previous actors to a new version and then copies the most basic migration, without any of the custom stuff that was needed for that last network upgrade. You can see that for v16 here: https://github.com/filecoin-project/go-state-types/pull/325

builtin/v9/migration/top.go
```diff
@@ -397,11 +397,12 @@ func MigrateStateTree(ctx context.Context, store cbor.IpldStore, newManifestCID
...
 					doneNow := doneCount
-					pendingNow := jobsNow - doneNow
 					elapsed := time.Since(startTime)
 					rate := float64(doneNow) / elapsed.Seconds()
-					log.Log(rt.INFO, "%d jobs created, %d done, %d pending after %v (%.0f/s)",
-						jobsNow, doneNow, pendingNow, elapsed, rate)
+					percentComplete := float64(doneNow) / float64(jobsNow) * 100
+
+					log.Log(rt.INFO, "Performing migration: %d of %d jobs complete (%.1f%%, %.0f/s)",
+						doneNow, jobsNow, percentComplete, rate)
```
rvagg, line 405: When you watch a migration happen using `lotus-shed` you might find that `percentComplete` is not helpful and we need to remove it. I hoped we could do this and my original issue I think said we should, but watching it progress myself we are adding jobs at the same time as completing them, so we never know what "complete" is. I don't think we have an easy way to pre-calculate the number of jobs. So, unfortunately, the main benefit of this logging is simply letting people know that it's happening and that their node isn't stuck.

virajbhartiya: Will something like this work 
```go
log.Log(rt.INFO, "Performing migration: %s of %s jobs processed (%s/s) [%v elapsed]", doneStr, jobsStr, rateStr, elapsed.Round(time.Second))
```
even though the `done` and `total` jobs will be changing, it will still show that the node isn't stuck and working just fine.

migration/runner.go
```diff
@@ -106,13 +107,17 @@ func RunMigration(ctx context.Context, cfg Config, cache MigrationCache, store c
...
 					rate := float64(doneNow) / elapsed.Seconds()
-					log.Log(rt.INFO, "%d jobs created, %d done, %d pending after %v (%.0f/s)",
-						jobsNow, doneNow, pendingNow, elapsed, rate)
+
+					jobsStr := fmt.Sprintf("%d", jobsNow)
+					doneStr := fmt.Sprintf("%d", doneNow)
+					rateStr := fmt.Sprintf("%.0f", rate)
+
+					log.Log(rt.INFO, "Performing migration: %s of %s jobs processed (%s/s) [%v elapsed]",
+						doneStr, jobsStr, rateStr, elapsed.Round(time.Second))
```
rvagg, line 120: PR is looking good @virajbhartiya but why do we have the separate `fmt.Sprintf`'s for the numbers and not just using the `log.Log` for that? If we're not pulling apart the `Log` call manually in lotus then they can be arbitrary values so numbers should be fine, no?

virajbhartiya: Oh right that could be done directly, thanks!, updated it

rvagg (approved):
nice, I'd like to see output before merging though

migration/runner.go
```diff
@@ -112,7 +110,7 @@ func RunMigration(ctx context.Context, cfg Config, cache MigrationCache, store c
 					rate := float64(doneNow) / elapsed.Seconds()
 
 					log.Log(rt.INFO, "Performing migration: %d of %d jobs processed (%.0f/s) [%v elapsed]",
-						doneNow, jobsNow, rate, elapsed.Round(time.Second))
+						doneNow, jobsNow, rate, elapsed.Round(100*time.Millisecond))
```
rvagg, line 113: let's leave this one as seconds, it's likely coming out on a seconds cadence anyway and nobody's using this for benchmarking

```suggestion
						doneNow, jobsNow, rate, elapsed.Round(time.Second))
```

rvagg (approved):
nice 👌

PR #333, opened 2024-12-10
Separate out claim check to to avoid double claim load in Lotus

Kubuxu:
Currently, F3 would iterate over claims AMT, then call the `MinerNominalPowerMeetsConsensusMinimum`, which would load the same AMT again.

### DIFF
--- a/builtin/v10/power/power_state.go
+++ b/builtin/v10/power/power_state.go
@@ -117,24 +117,8 @@ type CronEvent struct {
 	CallbackPayload []byte
 }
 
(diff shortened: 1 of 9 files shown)


rvagg (approved):
cool, you probably could have just done this for v15 and v16 though

Kubuxu:
> cool, you probably could have just done this for v15 and v16 though

I did it only for v16, `aider` did it for the others.

rvagg:
ok, that's pretty cool

PR #338, opened 2025-01-22
feat(miner): externalise SectorOnChainInfo from AMT

rvagg:
Ref: https://github.com/filecoin-project/builtin-actors/pull/1604

Draft for now, for benchmarking purposes

### DIFF
--- a/builtin/v15/miner/miner_types.go
+++ b/builtin/v15/miner/miner_types.go
@@ -317,6 +317,13 @@ func (sa Sectors) Get(sectorNumber abi.SectorNumber) (info *SectorOnChainInfo, f
 	return &res, true, nil
 }
 
+func (sa Sectors) ForEach(cb func(abi.SectorNumber, *SectorOnChainInfo) error) error {
+	var sector SectorOnChainInfo
+	return sa.Array.ForEach(&sector, func(i int64) error {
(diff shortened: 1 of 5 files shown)


rvagg:
Migration takes too long and I run out of memory for this

```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
1731013 lotus     20   0  414.0g 169.9g   5.4g S 380.7  90.4  91:32.50 lotus-shed
```

just before `Killed`, I don't have enough RAM to support this.

It got to:

```
2025-01-22T16:46:48.801+1100    INFO    fil-consensus   filcns/upgrades.go:3012 Performing migration: 1184187 of 1185201 jobs processed (749/s) [26m21s elapsed]
```

Which is `245000` miners processed (I had another log line I didn't push in this commit). So about 1/3rd of the way through in 26 minutes, on a fairly fast machine.

rvagg:
I think I'd like to explore an approach of updates to the AMT being externalised but existing sectors prior to the network upgrade left as they are. So we'd just have to do some byte sniffing - is it a CID, or is it a `SectorOnChainInfo`? I believe we've avoided such half-way schemas before, opting to do a full migration into a new schema.

PR #353, opened 2025-02-27
fip-0100: update the vesting funds types to the new layout

### LINKED ISSUE #350: FIP-0100: Update Miner VestingFunds

Stebalien:
I've also changed the `LoadVestingFunds` to return a `[]VestingFund` so we can abstract across the different underlying representations.

fixes #350


rvagg (approved):
nice that we shave off those enclsoing array bytes with the new transparent form too

Stebalien:
Gah. Wrong merge order.

PR #354, opened 2025-02-27
fip-0100: update the vesting funds types to the new layout

Stebalien:
I've also changed the `LoadVestingFunds` to return a `[]VestingFund` so we can abstract across the different underlying representations.

related to https://github.com/filecoin-project/builtin-actors/pull/1636


Stebalien:
Approved by @rvagg in https://github.com/filecoin-project/builtin-actors/pull/1636.

PR #359, opened 2025-03-05
fip-0100: sectoronchaininfo, deadline and expirationset type changes

rvagg:
Needs one of https://github.com/whyrusleeping/cbor-gen/pull/111 or https://github.com/whyrusleeping/cbor-gen/pull/112 first; and this isn't complete, I have ExpirationSet and Deadline serialization tests to add to match https://github.com/filecoin-project/builtin-actors/blob/481e18161e0af9a9a076327569c4e90fccf23fb6/actors/miner/tests/types_test.rs

### DIFF
--- a/builtin/v16/miner/cbor_gen.go
+++ b/builtin/v16/miner/cbor_gen.go
@@ -886,7 +886,7 @@ func (t *Deadlines) UnmarshalCBOR(r io.Reader) (err error) {
 	return nil
 }
 
-var lengthBufDeadline = []byte{139}
+var lengthBufDeadline = []byte{141}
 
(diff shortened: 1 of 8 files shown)


rvagg:
@Stebalien this is good to go but with the `big.Int{}` uninitialised. For now it's just documented where it matters that you need to take care touching that value. I think in practice that the only one that will be used (outside of invariants checks) is `SectorOnChainInfo.DailyFee` but it's going to be used in serialized form for the API and the serialisers already treat nil as zero so I think that should be fine. It might matter if miner software starts touching it though.

Stebalien (approved):
I'd say merge this as-is for now while we figure out the nil/zero stuff. LGTM.

PR #360, opened 2025-03-05
fix(big)!: treat uninitialized big.Int as zero

rvagg:


Stebalien:
I'd consider merging this change regardless (and, in that case, I'd merge it into master and rebase on it). We never actually want to panic if this is nil, and it's still possible to distinguish between nil and zero.

rvagg:
https://github.com/filecoin-project/lotus/pull/12936 has this pulled in to Lotus, looking at itest coverage for now.

rvagg:
ok, let's do this then

PR #361, opened 2025-03-06
fip-0100: add Deadline & Deadlines migration, summing LivePower

rvagg:


builtin/v16/migration/miner.go
```diff
@@ -76,6 +81,67 @@ func (m minerMigrator) MigrateState(ctx context.Context, store cbor.IpldStore, i
...
+
+		// Copy old deadline
+		newDeadline := miner16.Deadline{
+			Partitions:                        oldDeadline.Partitions,
+			ExpirationsEpochs:                 oldDeadline.ExpirationsEpochs,
+			PartitionsPoSted:                  oldDeadline.PartitionsPoSted,
+			EarlyTerminations:                 oldDeadline.EarlyTerminations,
+			LiveSectors:                       oldDeadline.LiveSectors,
+			TotalSectors:                      oldDeadline.TotalSectors,
+			FaultyPower:                       miner16.NewPowerPair(oldDeadline.FaultyPower.Raw, oldDeadline.FaultyPower.QA),
```
Stebalien, line 107: Ew. We really should move power pair into some form of shared type package.

Stebalien (approved):

builtin/v16/migration/miner.go
rvagg: Yes, I'm mentally making a list. I did Batch return already but we just need to do all the utility types that don't change.

PR #362, opened 2025-03-07
fip-0100: add daily fee state invariants

Stebalien:
1. Verify that the fees in the expiration queues match the fees associated with the sectors for each entry in each queue.
2. Verify that the sums of the fees in all the the expiration queues for each deadline match the fees in the deadlines.
3. Verify that the sum of the partition live power matches the live power in each deadline.


rvagg (approved):
nice, I didn't expect it to be so straightforward,

PR #397, opened 2025-07-04
feat: cbor encoding of all referenced param & return types

### LINKED ISSUE #396: Add cron params to codegen (🌽)
https://github.com/filecoin-project/go-state-types/blob/3cfb1b489520ab009985ac82557494cbe150becd/builtin/v17/gen/gen.go#L44

There are at least the ConstructorParams that we need to add. We should do this for all versions, probably do a quick check on other actors too to make sure we have it all covered.

rvagg:
Closes: https://github.com/filecoin-project/go-state-types/issues/396

`SectorContentChanged` is a little odd to include here, but it's referenced as in the market methods and uses the miner types for the call, which makes sense, but you don't typically even see it. Maybe if you were decoding internal calls between actors, or were doing some testing of a smart contract that uses it.

The *uint64 ones are because we have some methods that use them, `GetSectorSizeExported` and `GetDealSectorExported`. We don't really need the nullability, however, but we declare them as pointers in methods.go so we'll just be complete here.

### DIFF
--- a/abi/cbor_test.go
+++ b/abi/cbor_test.go
@@ -22,3 +22,45 @@ func TestCborBytesTransparent(t *testing.T) {
 	require.NoError(t, bytesDeSer.UnmarshalCBOR(bytes.NewReader(tBytes)))
 	require.Equal(t, tBytes, bytesDeSer)
 }
+
+func TestCborSectorSize(t *testing.T) {
+	tSectorSize := abi.SectorSize(20 * 1024)
+
+	buf := new(bytes.Buffer)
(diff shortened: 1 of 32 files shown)


rvagg:
OK, this fails because the codegen for my simple `transparent` type does this:

```go
func (t *PieceReturn) MarshalCBOR(w io.Writer) error {
	cw := cbg.NewCborWriter(w)

	// t.Accepted (bool) (bool)
	if err := cbg.WriteBool(w, t.Accepted); err != nil {
		return err
	}
	return nil
}
```

So that's a cbor-gen fix unfortunately.

PR #411, opened 2025-09-10
fix: `StateEncodeParams` API panic with Market Actor SectorContentChanged method

rjan90:
Fixes: https://github.com/filecoin-project/lotus/issues/13329

**Root Cause**
SectorContentChangedParams was defined as a type alias []SectorChanges, but CBOR-gen doesn't automatically generate marshalling methods for slice type aliases.

**Solution**
- Convert SectorContentChangedParams from type alias to proper type
- Add SectorContentChangedParams to gen.go files for automatic generation
- Regenerate CBOR marshalling methods using cbor-gen

### DIFF
--- a/builtin/v13/gen/gen.go
+++ b/builtin/v13/gen/gen.go
@@ -202,6 +202,7 @@ func main() {
 		miner.DataActivationNotification{},
 		miner.ProveReplicaUpdates3Params{},
 		miner.SectorUpdateManifest{},
+		miner.SectorContentChangedParams{},
 		miner.SectorChanges{},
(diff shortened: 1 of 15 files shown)


ZenGround0:
> Implement manual MarshalCBOR and UnmarshalCBOR methods 

@rjan90 did you use cbor gen here?  If not we should figure out how to get this done by cbor gen

ZenGround0:
Did you try adding the type [here](https://github.com/filecoin-project/go-state-types/blob/master/builtin/v17/gen/gen.go) ?  Is there an issue with following that approach?

Issue #409, opened 2025-09-05
Update method in market actor


akaladarshi:
- In [builtin-actor](https://github.com/filecoin-project/builtin-actors/blob/b7a35b4368377f090465553d0d6fe651fefb771b/actors/market/src/lib.rs#L1823C9-L1823C27) from v12 onwards we have replaced the `ActivateDeals` with the `BatchActivateDeals` method but in the [go-state-types](https://github.com/filecoin-project/go-state-types/blob/03987dc0ea56c28d1ee4c349e43fb029eddb4e82/builtin/v12/market/methods.go#L18) we are still using the old method (ActivateDeals).

- We need to update this method name from v12 till the latest market actor version

- Similarly the params this method accepts also needs to be updated:
  - [ActivateDealsParams](https://github.com/filecoin-project/go-state-types/blob/03987dc0ea56c28d1ee4c349e43fb029eddb4e82/builtin/v12/market/market_types.go#L45) need to be replaced with the [BatchActivateDealParams](https://github.com/filecoin-project/builtin-actors/blob/b7a35b4368377f090465553d0d6fe651fefb771b/actors/market/src/types.rs#L96), from v12 till latest

jintukumardas:
Hi @rvagg, could you assign this to me? I'd like to work on it.

jintukumardas:
@rjan90  Thank you for assigning it to me. I'll get back if I've any doubts.

PR #414, opened 2025-09-21
Replace ActivateDeals with BatchActivateDeals and update params (v12–latest)

jintukumardas:
Closes: #409

From builtin-actor v12 onwards, `ActivateDeals` was replaced with 
`BatchActivateDeals`, but go-state-types was still using the old method.

This change updates all affected versions (v12–latest) to align with the
current actor API.

- Method rename: ActivateDeals -> BatchActivateDeals
- Param type change: ActivateDealsParams -> BatchActivateDealsParams
- Updated parameter structure with sector-specific fields
- Regenerated CBOR for updated params

### DIFF
--- a/builtin/v12/gen/gen.go
+++ b/builtin/v12/gen/gen.go
@@ -117,7 +117,7 @@ func main() {
 		market.WithdrawBalanceParams{},
 		market.PublishStorageDealsParams{},
 		market.PublishStorageDealsReturn{},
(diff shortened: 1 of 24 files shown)


jintukumardas:
Hi @rjan90 , just checking if the changes look good now.

rvagg (approved):
cool, thanks!

PR #441, opened 2026-04-21
Include `v18` miner actors methods and types

akaladarshi:
Register the missing `v18` miner actor methods and their types:
 - GenerateSectorLocationExported
 - ValidateSectorStatusExported
 - GetNominalSectorExpirationExported


rjan90 (approved):
LGTM

